PR #7, opened 2015-01-13
HTML generators

### LINKED ISSUE #6: HTML Generators

jodosha:
## Introduction

For a design introduction of helpers please have a look at the [README](https://github.com/lotus/helpers/blob/master/README.md#design).

By including `HtmlHelper` it will inject **one private method:**  `html`.
This is a HTML5 markup generator.

## Usage

This is how it will look used with a view:

```ruby
class MyView
  include Lotus::Helpers::HtmlHelper

  # Generates
  # <aside id="sidebar">
  #    <div>hello</hello>
  #  </aside>
  def sidebar
    html.aside(id: 'sidebar') do
      div 'hello'
    end
  end
end
```

Features:
  * It knows how to close tags according to HTML5 spec (1)
  * It accepts content as first argument (2)
  * It accepts builder as first argument (3)
  * It accepts content as block which returns a string (4)
  * It accepts content as a block with nested markup builders (5)
  * It builds attributes from given hash (6)
  * it combines attributes and block (7)
  

```ruby
# 1
html.div # => <div></div>
html.img # => <img>

# 2
html.div('hello') # => <div>hello</div>

# 3
html.div(html.p('hello')) # => <div><p>hello</p></div>

# 4
html.div { 'hello' }
# =>
#<div>
#  hello
#</div>

# 5
html.div do
  p 'hello'
end
# =>
#<div>
#  <p>hello</p>
#</div>

# 6
html.div('hello', id: 'el', 'data-x': 'y') # => <div id="el" data-x="y">hello</div>

# 7
html.div(id: 'yay') { 'hello' }
# =>
#<div id="yay">
#  hello
#</div>
```

It supports complex scenarios.
**Notice the lack of concatenation between `div` and `input`** :sparkling_heart: 

```ruby
html.form(action: '/users', method: 'POST') do
  div do
    label 'First name', for: 'user-first-name'
    input type: 'text', id: 'user-first-name', name: 'user[first_name]', value: 'L'
  end

  input type: 'submit', value: 'Save changes'
end
# =>
#<form action="/users" method="POST">
#  <div>
#    <label for="user-first-name">First name</label>
#    <input type="text" id="user-first-name" name="user[first_name]" value="L">
#  </div>
#  <input type="submit" value="Save changes">
#</form>
```

## Custom tags

```ruby
html.tag(:custom, 'Foo', id: 'next') # => <custom id="next">Foo</custom>
html.empty_tag(:xr, id: 'next')      # => <xr id="next">
```

## Auto escape

The tag contents are automatically escaped:

```ruby
html.div('hello')         # => <div>hello</hello>
html.div { 'hello' }      # => <div>hello</hello>
html.div(html.p('hello')) # => <div><p>hello</p></hello>
html.div do
  p 'hello'
end # => <div><p>hello</p></hello>



html.div("<script>alert('xss')</script>")
  # =>  "<div>&lt;script&gt;alert(&apos;xss&apos;)&lt;&#x2F;script&gt;</div>"

html.div { "<script>alert('xss')</script>" }
  # =>  "<div>&lt;script&gt;alert(&apos;xss&apos;)&lt;&#x2F;script&gt;</div>"

html.div(html.p("<script>alert('xss')</script>"))
  # => "<div><p>&lt;script&gt;alert(&apos;xss&apos;)&lt;&#x2F;script&gt;</p></div>"

html.div do
  p "<script>alert('xss')</script>"
end
  # => "<div><p>&lt;script&gt;alert(&apos;xss&apos;)&lt;&#x2F;script&gt;</p></div>"
```

**HTML attributes aren't automatically escaped, developers must use: `Utils::Escape.html_attribute`, which will be aliased to `#ha` and make available as an helper in an upcoming PR**.

Ref https://github.com/lotus/utils/pull/51

## View Context

Local variables from views are available inside the nested blocks of HTML builder:

```ruby
require 'lotus/view'
require 'lotus/helpers'

Book = Struct.new(:title)

module Books
  class Show
    include Lotus::View
    include Lotus::Helpers::HtmlHelper

    def title_widget
      html.div do
        h1 book.title
      end
    end
  end
end
```

```erb
<div id="content">
  <%= title_widget %>
</div>
```

```ruby
book     = Book.new('The Work of Art in the Age of Mechanical Reproduction')
rendered = Books::Show.render(format: :html, book: book)
  # => <div id="content">
  #      <div>
  #        <h1>The Work of Art in the Age of Mechanical Reproduction</h1>
  #      </div>
  #   </div>
```

Closes #6


joneslee85:
@jodosha IMHO for unknown HTML tag, we should also provide an option to tell if the tag is self-closing or not

And besides, I don't see any form/input tag, are they going to be totally different?

lib/lotus/helpers/html_helper/html_builder.rb
```diff
@@ -0,0 +1,122 @@
+require 'lotus/helpers/html_helper/self_closing_html_node'
+require 'lotus/helpers/html_helper/html_node'
+
+module Lotus
+  module Helpers
+    module HtmlHelper
+      class HtmlBuilder < BasicObject
```
joneslee85: for the sake of clarity, maybe being explicit about `::BasicObject`

lib/lotus/helpers/html_helper/self_closing_html_node.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Helpers
+    module HtmlHelper
+      class SelfClosingHtmlNode
+        def initialize(name, attributes)
+          @name       = name
+          @attributes = attributes
+        end
+
+        def to_s
+          %(<#{ @name }#{attributes}>)
```
joneslee85: I could see that there is no closing `/` for this tag. So does this mean we only support HTML5? What about XHTML?

jodosha: Yes, HTML5 is the future.

joneslee85: Okay, fair enough. Since the generator generates `application.html` with tag `<!doctype HTML>`. I am happy with it.

jodosha:
@joneslee85 RE Forms: This PR adds the support for HTML building blocks. Once done we can think about forms. They involve not just markup, but also security, and knowledge of params/entities. ;)

joneslee85:
:+1:

thecatwasnot:
I cooked up a (small, very basic) [patch](https://gist.github.com/thecatwasnot/750a0c51fbe27b853060) to get the view context from the HtmlBuilder class, mostly because I was curious how it could be done ;)  It doesn't go looking for instance vars either, but that would be more heavy handed.  :+1: on the existing code, love the api it makes.

moonglum:
Hey :smile: Have you looked at @plexus' [hexp](https://github.com/plexus/hexp)? It could be used as a basis for something like this IMHO.

jodosha:
@moonglum I'm aware of hexp. However it will bring too much direct runtime dependencies:

  * concord
  * equalizer
  * nokogiri
  * sass

..and indirect:

  * adamantium
  * ice_nine
  * memoizable
  * thread_safe
  * mini_portile

It seems to be overkilling for a single feature, even if it's really important for this framework.

plexus:
Those are all tiny except nokogiri and SASS. Sass is only used for its css
selector parser, this should be replaced with the one from nokogiri. Think
that's the best I can do to trim dependencies.

plexus:
I didn't realize before that Hexp's dependencies were a blocker for you. I had a look at the numbers, these are source lines of code under `lib/`, excluding comment lines.

```
| hexp        | 1382 |
| concord     |   67 |
| equalizer   |   60 |
| adamantium  |  188 |
| ice_nine    |  312 |
| memoizable  |  222 |
| thread_safe | 1653 |
```

Adamantium pulls in the three below it, which I agree is a bit much. It does give you a level of immutable safety that you find nowhere else in Ruby, but that also very few care about. And as I've found out afterwards it adds a non-trivial cost to creating objects, I've thought about dropping it for that reason alone.

I can imagining splitting off a `hexp-core` gem which only contains the core data structures (`Node` and `List`), and the pure-ruby html generator. This would only have to depend on Concord and Equalizer. It could do shallow freezing instead of relying on Adamantium for deep freezing. Hypothetical `hexp-parse` and `hexp-selector` gems would depend on Nokogiri.

So with hexp-core you would get a pure-ruby virtual dom, builder syntax, and a html generator. No parsing of HTML to Hexp, no selecting nodes or manipulating trees based on CSS selectors.

This wouldn't even be that much work. In first instance hexp-selector could still depend on sass, making it almost purely an exercise of moving code around.

Hexp has seen very little adoption though, so I wonder if it's worth the effort. Maybe this is a chicken and egg situation? What do you think?

jodosha:
@plexus Yes, dependencies are always a blocking issue for Lotus. It isn't just a matter of lines of code, but it also means to make things faster for final users. They mean faster dev and CI setups, faster startup times (due to less Bundler overhead).

I would to say thank you :green_heart: :heart: :yellow_heart: :blue_heart: :heart: for your suggestion to split `hexp` in different gems for US. But, as you said, "I wonder if it's worth the effort".

> So with hexp-core you would get a pure-ruby virtual dom, builder syntax, and a html generator. No parsing of HTML to Hexp, no selecting nodes or manipulating trees based on CSS selectors.

Removing all that extra features you'll get more or less what this feature already implements. :wink: 

>  It does give you a level of immutable safety that you find nowhere else in Ruby, but that also very few care about.

I see the value of immutable data structures. However our scenario uses "throw-away" objects, and we don't keep references of them around. Probably, keeping them mutable, shouldn't bite us. :tiger: In Ruby, deep freezing objects is expensive. It would add an unnecessary overhead for this feature.

Those are the current LOCs count:

```
Utils:       423*
Helpers:      70**
Validations: 317
Router:      472
Model:       852
View:        498
Controller:  738
Lotus:       838

All the counts are related to the current repositories HEAD and measured with `simplecov` gem.

*  html-escape branch
** html-generators branch
```

Your library is super specialized to do one thing and do it probably better than `Lotus::Helpers`. At the same time, I feel confident that this PR provides a good enough feature to be shipped. :hamster:

plexus:
Alright, that sounds reasonable. :shipit:

moonglum:
:+1:

jodosha:
@thecatwasnot Hey, I've included a modified version of you patch. Thanks for the trick, you saved me a lot of time. :+1: 

I've discovered that `Binding` exposes the context with `#receiver`, so there is no need to use `Kernel.eval` to get the original view. :wink:

Gemfile
```diff
@@ -3,9 +3,12 @@ gemspec
 
 
 unless ENV['TRAVIS']
-  gem 'byebug', require: false, platforms: :mri if RUBY_VERSION >= '2.1.0'
+  gem 'byebug', require: false, platforms: :mri
   gem 'yard',   require: false
 end
 
+gem 'lotus-utils', github: 'lotus/utils', branch: 'master' # FIXME use a stable branch
```
AlfonsoUceda: careful, FIXME in the comment

lib/lotus/helpers/html_helper/empty_html_node.rb
```diff
@@ -0,0 +1,77 @@
...
+        private
+        # Resolve the attributes
+        #
+        # @return [String,NilClass] the tag attributes
+        #
+        # @since x.x.x
+        # @api private
+        def attributes
+          return if @attributes.nil?
+          result = [nil]
```
AlfonsoUceda: why this nil?

test/fixtures.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  def div_with_undescore_attr
+    html.div('data_where' => 'up')
+  end
+
+  def div_with_attrs
+    html.div(id: 'content', class: 'filled')
+  end
+
+  # FIXME add more cases from:
```
AlfonsoUceda: This is done?

AlfonsoUceda:
This is perfect @jodosha beautiful job :heart: :clap:  :tophat:

lib/lotus/helpers/html_helper/empty_html_node.rb
jodosha: In case there are attributes we want to leave a space between the name of the tag and the attributes:

```ruby
name = :hr
attributes = ([nil] << ['id="foo"', 'class="separator"']).join(' ')

%(<#{ name }#{attributes}>) # => <hr id="foo" class="separator">
```

AlfonsoUceda: and why not...

```ruby
name = :hr
attributes = ['id="foo"', 'class="separator"'].join(' ')

%(<#{ name } #{attributes}>)
```
maybe I'm missing something

test/fixtures.rb
jodosha: Well, that was a note from the initial implementation, before we introduced `Utils::Escape`. Specifically, that is about escaping tag attributes.

We don't do it automatically, because it's technically impossible, but we have `Utils::Escape.html_attribute` which will be available soon as `#ha` helper.

The technical limitation mentioned above is about the impossibility to distinguish between what should be escaped and what not. Consider the following example:

```ruby
def home_page_link
  html.a('home', href: '/')
end

home_page_link # => <a href="&#x2f;">home</a> # WRONG
```

Developers should escape attributes by themselves only if those contain user inputs.

lib/lotus/helpers/html_helper/empty_html_node.rb
jodosha: In case attributes are not present, the output would be `<hr >` or `<div ></div>`. Notice the blank space between the last letter of the name and the closing tag character.

AlfonsoUceda: ok, then perfect :P

lib/lotus/helpers/html_helper/html_builder.rb
```diff
@@ -0,0 +1,288 @@
...
+        #
+        # @return [Lotus::Helpers::HtmlHelper::HtmlBuilder] the builder
+        #
+        # @since x.x.x
+        # @api private
+        def initialize
+          @nodes = []
+        end
+
+        # Defines a custom tag
```
joneslee85: /s/Defines/Define

joneslee85:
:+1:

PR #11, opened 2015-02-08
Number formatting helper

### LINKED ISSUE #3: Number formatting
Introduce a module that provides formatting facilities for numbers.
It should accept any kind of number and return a comma delimited string, with decimal separator.

Please consider to include both unit and integration tests. Please have a look at `lotus-assets` for integration tests example.

The goal is to provide private methods (if possible). This should forbid bad usage of those helpers:

```erb
<%= format_number cart.total %>
```

The aimed style is to let templates to bind only one expression.

```erb
<%= total %>
```

Where `#total` is defined like this:

```ruby
class MyView
  include Lotus::Helpers::Numbers

  def initialize(cart)
    @cart = cart
  end

  def total
    format_number @cart.total
  end
end
```

tomkadwill:
Number formatting helper as discussed in https://github.com/lotus/helpers/issues/3

Closes #3

### DIFF
--- a/lib/lotus/helpers/numbers.rb
+++ b/lib/lotus/helpers/numbers.rb
@@ -47,8 +47,8 @@ module Lotus
       #
       #   view.euros
       #     # => "1.256,95"
-      def format_number(number, delimiter: ',', separator: '.', precision: nil)
-        number_parser = NumberFormatter.new(number, delimiter, separator, precision)
+      def format_number(number, options = {})
+        number_parser = NumberFormatter.new(number, options)
         number_parser.parse
       end
     end
--- a/lib/lotus/helpers/numbers.rb
+++ b/lib/lotus/helpers/numbers.rb
@@ -23,20 +23,18 @@ module Lotus
       #
       # @since x.x.x
       #
-      # @see Lotus::Helpers::Numbers#format_number
-      #
       # @example Basic usage
       #   require 'lotus/helpers/numbers'
       #
-      #   class checkout
+      #   class Checkout
       #     include Lotus::Helpers::Numbers
       #
       #     def total
-      #       n 1999.99
+      #       format_number 1999.99
       #     end
       #
       #     def euros
-      #       n 1256.95, delimiter: '.', separator: ','
+      #       format_number 1256.95, delimiter: '.', separator: ','
       #     end
       #   end
       #
--- a/lib/lotus/helpers/numbers.rb
+++ b/lib/lotus/helpers/numbers.rb
@@ -1,0 +1,56 @@
+require 'lotus/helpers/numbers/number_formatter'
+
+module Lotus
+  module Helpers
+    # Number formatter
+    #
+    # You can include this module inside your view and
+    # the view will have access all methods.
+    #
+    # By including <tt>Lotus::Helpers::Numbers</tt> it will inject private
+    # methods as markup escape utilities.
+    #
+    # @since x.x.x
+    module Numbers
+      private
+      # Number formatting
+      #
+      # This helper formats raw numbers to make them more human readable
+      #
+      # @param number [Numeric] the number to be formatted
+      #
+      # @return [String] formatted number
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Helpers::Numbers#format_number
+      #
+      # @example Basic usage
+      #   require 'lotus/helpers/numbers'
+      #
+      #   class checkout
+      #     include Lotus::Helpers::Numbers
+      #
+      #     def total
+      #       n 1999.99
+      #     end
+      #
+      #     def euros
+      #       n 1256.95, delimiter: '.', separator: ','
+      #     end
+      #   end
+      #
+      #   view = MyView.new
+      #
+      #   view.total
+      #     # => "1,999.99"
+      #
+      #   view.euros
+      #     # => "1.256,95"
+      def format_number(number, delimiter: ',', separator: '.', precision: nil)
+        number_parser = NumberFormatter.new(number, delimiter, separator, precision)
+        number_parser.parse
+      end
+    end
+  end
+end
--- a/lib/lotus/helpers/numbers/number_formatter.rb
+++ b/lib/lotus/helpers/numbers/number_formatter.rb
@@ -21,11 +21,11 @@ module Lotus
 
         attr_accessor :number, :delimiter, :separator, :precision
 
-        def initialize(number, delimiter, separator, precision)
+        def initialize(number, options)
           @number = number
-          @delimiter = delimiter
-          @separator = separator
-          @precision = precision
+          @delimiter = options.fetch(:delimiter, ',')
+          @separator = options.fetch(:separator, '.')
+          @precision = options.fetch(:precision, nil)
         end
 
         def parts
--- a/lib/lotus/helpers/numbers/number_formatter.rb
+++ b/lib/lotus/helpers/numbers/number_formatter.rb
@@ -16,7 +16,7 @@ module Lotus
 
         private
 
-        DELIMITED_REGEX = /(\d)(?=(\d\d\d)+(?!\d))/
+        REGEX_TO_DELIMIT = /(\d)(?=(\d{3})+$)/
         DEFAULT_SEPARATOR = '.'
 
         attr_accessor :number, :delimiter, :separator, :precision
@@ -29,15 +29,16 @@ module Lotus
         end
 
         def parts
-          left, right = split_number
-          left.gsub!(DELIMITED_REGEX) do |digit_to_delimit|
-            "#{digit_to_delimit}#{delimiter}"
-          end
-          [left, right].compact
+          first, second = split_number
+          [add_delimiter_to(first), second].compact
         end
 
         def split_number
           to_str.split(DEFAULT_SEPARATOR)
+        end
+
+        def add_delimiter_to(n)
+          n.gsub(REGEX_TO_DELIMIT) { |digit| "#{digit}#{delimiter}" }
         end
 
         def rounded_number
@@ -53,8 +54,6 @@ module Lotus
         def to_f
           case rounded_number
           when NilClass
-            raise TypeError
-          when String
             raise TypeError
           when Fixnum
             number
--- a/lib/lotus/helpers/numbers/number_formatter.rb
+++ b/lib/lotus/helpers/numbers/number_formatter.rb
@@ -17,6 +17,7 @@ module Lotus
         private
 
         DELIMITED_REGEX = /(\d)(?=(\d\d\d)+(?!\d))/
+        DEFAULT_SEPARATOR = '.'
 
         attr_accessor :number, :delimiter, :separator, :precision
 
@@ -36,8 +37,7 @@ module Lotus
         end
 
         def split_number
-          type_check!
-          rounded_number.to_s.split('.')
+          to_str.split(DEFAULT_SEPARATOR)
         end
 
         def rounded_number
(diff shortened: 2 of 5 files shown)


lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,17 @@
...
+module Lotus
+  module Helpers
+    # TODO
+    #
+    module Numbers
+      private
+      # TODO
+      #
+      def format_number(number, options = {})
+        number_parser = NumberParser.new(number, options)
```
jodosha: Why the implementation of `format_number` is delegated to `NumberParser`. Isn't better to call it `NumberFormatter`?

jodosha:
@tomkadwill Thanks for this PR, can you please add logic for rounding too?

lib/lotus/helpers/numbers.rb
tomkadwill: Yes that makes more sense, I'll rename the class.

tomkadwill:
No problem. Yes, I assume this will be an extra option? e.g.:

```
format_number 5002.007, precision 1
# => 5,002.00
```

jodosha:
Yes for `precision`. Default 2 digits, but it should only present those for non integer numbers.

```ruby
format_number 10 # => 10
format_number 3.141592 # => 3.14

format_number 10, precision: 4 # => 10.0000
format_number 3.141592, precision: 4 # => 3.1415
```

tomkadwill:
@jodosha Implemented changes. Now hopefully ready to merge :smile:

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,59 @@
...
+      #   end
+      #
+      #   view = MyView.new
+      #
+      #   view.total
+      #     # => "1,999.99"
+      #
+      #   view.euros
+      #     # => "1.256,95"
+      def format_number(number, options = {})
```
jodosha: @tomkadwill I think this can be a good candidate for keyword arguments. Would you mind to try this approach?
It will help to define inline defaults and leave `NumberFormatter` to not decide those.

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,59 @@
...
+        number_parser = NumberFormatter.new(number, options)
+        number_parser.parse
+      end
+
+      # @since x.x.x
+      alias_method :n, :format_number
```
jodosha: We have defined `#h` et co because it's a well known idea in Ruby web frameworks, and because we want to have less noise in HTML generator.

But we should limit as much as possible to overload views contexts. What do you think if we remove it for now?

lib/lotus/helpers/numbers/number_formatter.rb
```diff
@@ -0,0 +1,52 @@
...
+      # Numbers
+      #
+      # @since x.x.x
+      class NumberFormatter
+        # NumberFormatter
+        #
+        # @since x.x.x
+        # @api private
+
+        attr_accessor :number, :delimiter, :separator, :precision
```
jodosha: If there isn't a good reason to keep those public, should we declare those as private?

lib/lotus/helpers/numbers.rb
tomkadwill: @jodosha Sure I will give that a try. However, would this be problematic in the future if the number of keyword arguments becomes large?

lib/lotus/helpers/numbers.rb
tomkadwill: Yes, no problem, I will remove

lib/lotus/helpers/numbers/number_formatter.rb
tomkadwill: @jodosha I can make make `initialize` and the `attr_accessors` private but the `parse` method needs to be public. Is that what you mean?

jodosha: Yes. :+1:

lib/lotus/helpers/numbers.rb
jodosha: @tomkadwill Let's see this an opportunity to avoid too much args :wink: I think the actual set covers it all. The only thing that we may pass in the future is a locale code.

tomkadwill:
@jodosha Update PR and rebased :)

jodosha:
@tomkadwill Hey thanks for the quicks iterations. :hand: 

There are a few other behaviors that worth to be decided and then tested:

  * What happens if `nil` is passed? Raise `TypeError`?
  * What happens if a non numeric type is passed? Raise `TypeError`?
  * What happens if a string representing a number is passed? Eg. `format_number '23'` and what about `format_number '23.10'`?
  * What happens with `BigDecimal`, `Complex` and `Rational`?
  * Do we want to support explicit protocol conversions? Eg. `#to_int`, `#to_f`, `#to_d`?

tomkadwill:
Yep those are all good points, I would say:

* If `nil` is passed as the first param: raise `TypeError`.
* If non numeric type is passed as first param: raise `TypeError`.
* If string representing number is passed: We could try calling `to_f` or `to_i` on strings, however, that might bloat the helper. Perhaps we should just raise a `TypeError` and only accept numbers?
*  In the case of `BigDecimal`, `Complex` and `Rational` it feels like we should try and convert to float? and perhaps raise an error if the conversion is not possible?
* Can you explain what you mean by 'supporting explicit protocol conversions'?

cc @jodosha

jodosha:
> Can you explain what you mean by 'supporting explicit protocol conversions'?

Code example: https://github.com/lotus/utils/blob/master/test/kernel_test.rb#L564

tomkadwill:
Ah, ok, thanks for the clarification :smile: 

 I would argue that supporting explicit protocol conversions is a feature that could be added at a later date without breaking the public API. What do you think? And what do you think about the points in my last comment?

jodosha:
@tomkadwill I agree with all your comments :+1:.

For converting a number to `Float` or raise a `TypeError` you can use `Utils::Kernel`.

tomkadwill:
@jodosha Thanks for the feedback and pointing me to `Utils::Kernel.Float`!

I've updated the PR to cater for the scenarios that we discussed. Awaiting your feedback :blush:

jodosha:
@tomkadwill I think that all the type checking can be simplified with:

```ruby
private
def to_f
  case @number
    when NilClass
       raise TypeError
    else
      Utils::Kernel.Float(@number)
    end
  end
end

def to_str
  to_f.to_s
end

def split_number
  to_str.split(DEFAULT_SEPARATOR)
end
```

**I haven't tried locally, but I think it should work**. Please consider this refactoring. Thank you! :)
This will solve a lot of cases:
  * `nil` will raise `TypeError`
  * `Integer` will convert to a float, but then we can reject the decimal part.
  * `Float` will be returned as it was.
  * `BigDecimal`, `Complex` and `Rational` will be converted into a float for presentational purposes.
  * String representing numbers will be coerced into floats
  * All the other strings will raise `TypeError`

AlfonsoUceda:
@tomkadwill why removed keywords arguments?

tomkadwill:
@AlfonsoUceda originally I wrote the code to be compatible with Ruby 2.2. @jodosha asked my to change my commit to be compatible with versions < 2.2.

Waiting until 0.1.0 ships before merging

jodosha:
@tomkadwill I was tidying up the code when I recognized that this [implementation is from Rails](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/number_helper/number_to_delimited_converter.rb) codebase without any acknowledgement.

This is unacceptable for me. If you still want take care of this you should solve this issue with one of the following solutions:

  1. Create an alternative implementation (preferred)
  2. Add a copyright note that the code is taken from Rails.

The option 1 is preferred because this PR is still incomplete (sigh). We should support strings. For instance params represent everything like a string. This will be a common use case.

Can you please rework and expand? Thank you.

tomkadwill:
@jodosha You are right. I did copy some of this code from the Rails implementation. To be honest, I forgot to add an acknowledgement and that was not good. I didn't mean to behave unethically.

I'll rework the solution or add acknowledgement where appropriate.

joneslee85:
@tomkadwill ping, ready for review yet?

tomkadwill:
@joneslee85 Yep ready for review. Sorry, I sent @jodosha a message on gitter.im to review but I forgot to update the ticket here.

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,56 @@
...
+      # @return [String] formatted number
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Helpers::Numbers#format_number
+      #
+      # @example Basic usage
+      #   require 'lotus/helpers/numbers'
+      #
+      #   class checkout
```
AlfonsoUceda: s/checkout/Checkout

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,56 @@
...
+      #
+      # This helper formats raw numbers to make them more human readable
+      #
+      # @param number [Numeric] the number to be formatted
+      #
+      # @return [String] formatted number
+      #
+      # @since x.x.x
+      #
+      # @see Lotus::Helpers::Numbers#format_number
```
AlfonsoUceda: remove this line because it is the same method

lib/lotus/helpers/numbers/number_formatter.rb
```diff
@@ -0,0 +1,72 @@
+module Lotus
+  module Helpers
+    module Numbers
+      # Numbers
+      #
+      # @since x.x.x
+      class NumberFormatter
```
AlfonsoUceda: This class should be commented

lib/lotus/helpers/numbers/number_formatter.rb
```diff
@@ -0,0 +1,72 @@
...
+        # NumberFormatter
+        #
+        # @since x.x.x
+        # @api private
+
+        def parse
+          parts.join(separator)
+        end
+
+        private
```
AlfonsoUceda: I think this `private` can be removed, what do you think?

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,56 @@
...
+      #
+      # @example Basic usage
+      #   require 'lotus/helpers/numbers'
+      #
+      #   class checkout
+      #     include Lotus::Helpers::Numbers
+      #
+      #     def total
+      #       n 1999.99
```
AlfonsoUceda: s/n/format_number

lib/lotus/helpers/numbers.rb
```diff
@@ -0,0 +1,56 @@
...
+      #     end
+      #
+      #     def euros
+      #       n 1256.95, delimiter: '.', separator: ','
```
AlfonsoUceda: s/n/format_number

lib/lotus/helpers/numbers/number_formatter.rb
tomkadwill: @AlfonsoUceda Thanks for your comments. I didn't comment this class because I thought of it as a 'private' class that is only used internally. Therefore, it shouldn't appear in the public documentation.

lib/lotus/helpers/numbers/number_formatter.rb
tomkadwill: @AlfonsoUceda I made these methods private as they are only used internally. Why do you think it should be removed?

joneslee85:
@jodosha @AlfonsoUceda let's get this reviewed and merge :)

jodosha:
@tomkadwill I've just merged this. Thank you! :+1:

PR #14, opened 2015-02-12
Routing helper

jodosha:
### What
Routing helper for full stack Lotus web applications.

### Explanation
For a given application called `Web::Application`, at the runtime Lotus creates a routes factory called `Web::Routes`.

By including this module in a view, it makes that factory avaliable as `routes`.

### Basic usage in template

```ruby
module Web::Views::Home
  class Index
    include Web::View
    include Lotus::Helpers
  end
end

# ERB template
# <%= routes.home_path %>
```

### Basic usage in view

```ruby
module Web::Views::Home
  class Index
    include Web::View
    include Lotus::Helpers
  end

  def link_to_home
    %(<a href="#{ routes.home_path }">Home</a>)
  end
end

# ERB template
# <%= link_to_home %>
```

### Technical notes

This is different from Rails routing helpers. Over there, they are available directly in templates and helpers. Here, I wanted to avoid to overload views with a lot of method missing hackery.

Right now views respond to:

  * concrete methods defined by developers
  * exposures from actions
  * helpers

Adding all the routes available for an application as direct methods would increase a lot the complexity of the view scope. It's better to have the `routes` factory even if a little bit unhandy.

Ref https://github.com/lotus/lotus/pull/151


lib/lotus/helpers/routing_helper.rb
```diff
@@ -0,0 +1,52 @@
+module Lotus
+  module Helpers
+    # Routing helper for full stack Lotus web applications.
+    #
+    # For a given application called <tt>Web::Application</tt>, at the runtime
```
tomkadwill: 'at runtime' instead of 'at the runtime'

joneslee85:
:100:

PR #18, opened 2015-03-11
Add link_to helper

tomkadwill:
LinkTo Helper
===========

Provides a helper method for generating links. As dicussed on [discuss.lotusrb.org #58](https://discuss.lotusrb.org/t/link-to-view-helper/58)

#### Example Usage

    link_to 'blogs', :blogs_path
    link_to 'archives', '/archives', class: archives

### DIFF
--- a/lib/lotus/helpers/link_to_helper.rb
+++ b/lib/lotus/helpers/link_to_helper.rb
@@ -1,0 +1,28 @@
+require 'lotus/helpers/html_helper'
+
+module Lotus
+  module Helpers
+    # LinkTo Helper
+    #
+    # Including <tt>Lotus::Helpers::LinkTo</tt> will include the `link_to` method.
+    #
+    # @since x.x.x
+    #
+    # @see Lotus::Helpers::HtmlHelper#html
+    #
+    # @example Usage
+    #   # 1
+    #   link_to('home', '/') # => <a href="/">home</a>
+    #
+    #   # 2
+    #   link_to('users', :users_path, class: 'my_users') # => <a href="/users" class="my_users">users</div>
+    module LinkToHelper
+      include Lotus::Helpers::HtmlHelper
+
+      def link_to(content, url, options = {}, &blk)
+        options[:href] = url
+        html.a(blk || content, options).to_s
+      end
+    end
+  end
+end
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -232,11 +232,7 @@ class LinkTo
 
   class Routes
     def self.path(name, id = nil)
-      if id
-        "/#{ name }/#{id}"
-      else
-        "/#{ name }"
-      end
+      "/#{name}" << "/#{id}"
     end
   end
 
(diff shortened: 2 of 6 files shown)


test/fixtures.rb
```diff
@@ -226,3 +226,38 @@ def routing_helper_path
...
+  def link_to_post
+    link_to('Post', routes.path(:post, 1))
+  end
+
+  def link_to_with_class
+    link_to('Post', routes.path(:posts), class: 'first')
+  end
+
+  def link_to_with_id
+    link_to('Post', routes.path(:posts), id: 'posts_link')
```
joneslee85: What do you think about changing `posts_link` to BEM syle `posts__link`? Pleaser read more at http://www.sitepoint.com/bem-smacss-advice-from-developers/ - It's for the sake of encouraging good frontend practices

joneslee85:
:+1:

tomkadwill:
@joneslee85 sure. I've changed the test to BEM style

joneslee85:
:+1:

test/fixtures.rb
```diff
@@ -226,3 +226,38 @@ def routing_helper_path
     routes.path(:dashboard)
   end
 end
+
+class LinkTo
+  include Lotus::Helpers::LinkToHelper
+
+  class Routes
+    def self.path(name, id = nil)
```
AlfonsoUceda: in this method you can use "/#{name}" << "/#{id}"
```ruby
"/#{name}" << "/#{1}"
# => "/posts/1"

"/#{name}" << "/#{nil}"
# => "/posts/"
```

tomkadwill:
Thanks @AlfonsoUceda. I've updated the tests

lib/lotus/helpers/link_to_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+    #   link_to('home', '/') # => <a href="/">home</a>
+    #
+    #   # 2
+    #   link_to('users', :users_path, class: 'my_users') # => <a href="/users" class="my_users">users</div>
+    module LinkToHelper
+      include Lotus::Helpers::HtmlHelper
+
+      def link_to(content, url, options = {}, &blk)
+        options[:href] = url
+        html.a(blk || content, options).to_s
```
joneslee85: do we need to cast `to_s`? what do you think about leaving it to the template to do that job? Most of the cases the helper would be called within `<%= %>` block anyway, which cast `to_s`.

jodosha: @joneslee85 Good, catch! `#to_s` is better, so the public API of this will always return a String.

jodosha:
@tomkadwill This is good.

It still lacks of an integration test. The reason is simple: developers will use helpers in views. Views have autoescape, and we are generating HTML here, so we want to ensure that the browser will render a link, not escaped markup.

Can you please add it to this PR? There are examples already in place, where to take inspiration form. Thank you!

tomkadwill:
@jodosha I've added an integration test to test the link_to helper in a view. Is that what you were after?

joneslee85:
@tomkadwill :+1:

PR #22, opened 2015-07-13
Improve link_to with html content param

vyper:
Improve support for html content in link_to helper

# Before
```ruby
link_to(nil, routes.path(:posts)) do
  strong 'Post'
end
# => <a href="/posts/"><strong>Post</strong></a>
```

# Now
```ruby
link_to(routes.path(:posts)) do
  strong 'Post'
end
# => <a href="/posts/"><strong>Post</strong></a>
```

### DIFF
--- a/lib/lotus/helpers/link_to_helper.rb
+++ b/lib/lotus/helpers/link_to_helper.rb
@@ -9,17 +9,49 @@ module Lotus
     # @since x.x.x
     #
     # @see Lotus::Helpers::HtmlHelper#html
-    #
-    # @example Usage
-    #   # 1
-    #   link_to('home', '/') # => <a href="/">home</a>
-    #
-    #   # 2
-    #   link_to('users', :users_path, class: 'my_users') # => <a href="/users" class="my_users">users</div>
     module LinkToHelper
       include Lotus::Helpers::HtmlHelper
 
-      def link_to(content, url, options = {}, &blk)
+      # Link to url using content or html block
+      #
+      # @overload link_to(content, url, options)
+      #   Use string as content
+      #   @param content [String] content used in the a tag
+      #   @param url [String] url used in href attribute
+      #   @param options [Hash] HTML attributes to pass to the a tag
+      #
+      # @overload link_to(url, options, &blk)
+      #   Use block as content
+      #   @param url [String] url used in href attribute
+      #   @param options [Hash] HTML attributes to pass to the a tag
+      #   @param blk [Proc] A block that describes the contents of the a tag
+      #
+      # @return [String] the a tag
+      #
+      # @since x.x.x
+      #
+      # @example Usage
+      #   # 1
+      #   link_to('home', '/') # => <a href="/">home</a>
+      #
+      #   # 2
+      #   link_to('users', :users_path, class: 'my_users') # => <a href="/users" class="my_users">users</div>
+      #
+      #   # 3
+      #   link_to(:users_path) do
+      #     strong 'Users'
+      #   end # => <a href="/users"><strong>Users</strong></a>
+      def link_to(*args, **options, &blk)
+        if block_given? && args.size != 1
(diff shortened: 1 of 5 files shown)


AlfonsoUceda:
I like to me, what do you think @jodosha @joneslee85

joneslee85:
I like it too but I am questioning about what if the user provide both the param and the block?

AlfonsoUceda:
then I think it crash :S nice catch. @vyper please could you confirm this?

vyper:
@AlfonsoUceda and @joneslee85: Really it would crash, but I changed implementation and I improved documentation.

joneslee85:
@vyper :+1:

jodosha:
@vyper I like the idea and thank you for this PR.
There is one behavior to change IMO, which is the one that accepts both `content` and `blk`. From my understanding the former gets ignored. This can be source of confusion for devs. We should instead detect and handle two cases with an `ArgumentError` exception.

  1. When both `content` and `blk` are specified (`link_to('foo', routes.path(:home)) { 'bar' }`)
  2. When none of them is specified (`link_to()`).

There is also a third case that can be confusing:

```ruby
<%= link_to('foo') { 'bar' } %>
```

In unit tests it returns:

```html
<a href="foo">
bar
</a>
```

Please note that's an invalid `href`attribute. If you specify `/foo`, our unit tests return the expected value, but if you use in combination with Lotus::View, its security feature will autoescape `/`.

Having such a liberal signature like `def link_to(*content, **args, &blk)` may lead to unexpected usages that can lead devs to confusion.

@vyper Can you please add tests for the behaviors above? At this point it's clear we also need to expand integration tests (see `test/integration/link_to_helper_test.rb`). And as a last thing, when we're sure that the tests cover these edge cases, please experiment with a different signature. It would be great to have a stricter version of it. Thank you very much.

vyper:
@jodosha better that way?

lib/lotus/helpers/link_to_helper.rb
```diff
@@ -9,17 +9,49 @@ module Helpers
...
+      #   end # => <a href="/users"><strong>Users</strong></a>
+      def link_to(*args, **options, &blk)
+        if block_given? && args.size != 1
+          raise ArgumentError, 'Only an uri is expected as an argument when a block is given'
+        end
+
+        if !block_given? && args.size != 2
+          raise ArgumentError, 'Both content and url are expected as arguments'
+        end
+
```
joneslee85: what happen if there are more than 3 arguments?

vyper: raises exception `Both content and url are expected as arguments` when without block and `Only an uri is expected as an argument when a block is given` with block.

test/link_to_helper_test.rb
```diff
@@ -20,4 +20,32 @@
...
+
+  it 'returns a link with html content' do
+    @view.link_to_with_html_content.must_equal %(<a href="/posts/">\n<strong>Post</strong>\n</a>)
+  end
+
+  it 'returns a link with html content, id and class' do
+    @view.link_to_with_html_content_id_and_class.must_equal %(<a id="posts__link" class="first" href="/posts/">\n<strong>Post</strong>\n</a>)
+  end
+
+  it 'raises an link with content and html content' do
```
joneslee85: s/an/a

joneslee85: raises Error not a link

joneslee85:
@vyper looks good now, please squash into one and I'll get it merge

vyper:
@joneslee85 done. (:

jodosha:
@vyper @joneslee85 I've implemented a stricter signature in https://github.com/lotus/helpers/commit/153b65977e1fde416ef18826c037f5cbd7f60bb4, by keeping the same features.

joneslee85:
@jodosha your implementation :+1:

PR #26, opened 2015-07-27
Added support for boolean attributes.

pascalbetz:
Boolean attributes are rendered differently than normal attributes.
If its value is _false_ then the whole attribute is omitted.
If its value is _true_ then the attribute is rendered with its name as the value.

+ added special handling of boolean attributes
+ added tests
! fixed existing tests that checked for „true“ values in rendered boolean attributes

### DIFF
--- a/lib/lotus/helpers/html_helper/empty_html_node.rb
+++ b/lib/lotus/helpers/html_helper/empty_html_node.rb
@@ -6,6 +6,20 @@ module Lotus
       # @since 0.1.0
       # @api private
       class EmptyHtmlNode
+        # List of attributes that get special treatment when rendering.
+        # @see http://www.w3.org/html/wg/drafts/html/master/infrastructure.html#boolean-attribute
+        # @since x.x.x
+        # @api private
+        BOOLEAN_ATTRIBUTES = %w{allowfullscreen async autobuffer autofocus
+          autoplay checked compact controls declare default defaultchecked
+          defaultmuted defaultselected defer disabled draggable enabled
+          formnovalidate hidden indeterminate inert ismap itemscope loop
+          multiple muted nohref noresize noshade novalidate nowrap open
+          pauseonexit pubdate readonly required reversed scoped seamless
+          selected sortable spellcheck translate truespeed typemustmatch
+          visible
+        }.freeze
+
         # Attributes separator
         #
         # @since 0.1.0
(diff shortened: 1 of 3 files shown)


lib/lotus/helpers/html_helper/empty_html_node.rb
```diff
@@ -6,6 +6,12 @@ module HtmlHelper
       # @since 0.1.0
       # @api private
       class EmptyHtmlNode
+        # List of attributes that get special treatment when rendering.
+        # @see @see http://www.w3.org/html/wg/drafts/html/master/infrastructure.html#boolean-attribute
```
joneslee85: remove dup `@see`

lib/lotus/helpers/html_helper/empty_html_node.rb
```diff
@@ -6,6 +6,12 @@ module HtmlHelper
...
+        # @since 0.2.x
+        # @api private
+        BOOLEAN_ATTRIBUTES = %w{allowfullscreen async autobuffer autofocus autoplay checked compact controls declare default defaultchecked defaultmuted defaultselected defer disabled draggable enabled formnovalidate hidden indeterminate inert ismap itemscope loop multiple muted nohref noresize noshade novalidate nowrap open pauseonexit pubdate readonly required reversed scoped seamless selected sortable spellcheck translate truespeed typemustmatch visible}.freeze
```
joneslee85: can we move them to next line to fit all into 120 chars per line?

pascalbetz:
@joneslee85 please review, included your feedback and squashed the commits.

lib/lotus/helpers/html_helper/empty_html_node.rb
```diff
@@ -6,6 +6,20 @@ module HtmlHelper
       # @since 0.1.0
       # @api private
       class EmptyHtmlNode
+        # List of attributes that get special treatment when rendering.
+        # @see http://www.w3.org/html/wg/drafts/html/master/infrastructure.html#boolean-attribute
+        # @since 0.2.?
```
joneslee85: please leave it as x.x.x for us

lib/lotus/helpers/html_helper/empty_html_node.rb
```diff
@@ -45,13 +59,31 @@ def to_s
...
+          result
+        end
+
+        def boolean_attribute?(attribute_name)
+          BOOLEAN_ATTRIBUTES.include?(attribute_name.to_s)
+        end
+
+        # Do not render boolean attributes when their value is _false_.
+        def boolean_attribute(attribute_name, value)
+          return '' if !value
```
joneslee85: what if users provide value as text, such as `true`? What do you think about checking this value against `true` instead? Furthermore, I like to see this `if` inside `boolean_attribute?`

joneslee85: I would do like this:

```

Comment (pascalbetz) [inline lib/lotus/helpers/html_helper/empty_html_node.rb]: Thanks for the feedback.

We need to detect 3 cases: 

* boolean attribute and value false: ommit
* boolean attribute and value true: render boolean attribute
* normal attribute

So we can not move value comparison to the boolean_attribute? method (we would need to check if it is a boolean attribute with false value again).

What we could do though is moving the "next" inside the main if as statement modifier or as next. Here a sample with statement modifier:

```ruby
if boolean_attribute?(attribute_name)
  result << boolean_attribute(attribute_name, value) if value
else
 ...
end
```
As for the comparison: i'd use ruby _truthiness_ (everything but nil and false is true) instead of an explicit comparison to true. This is common behavior (not sure how lotus treats it).

What do you think.

joneslee85: sound good, go for it

pascalbetz:
@joneslee85 

Done, rebased against master, squashed.

Issue #29, opened 2015-08-14
Support for text nodes


josemota:
I was making my way towards building a form using the HTML5 API. However, it seems to be that it's currently impossible to accomplish the following:

```html
<p>
  <label>
    Your name?
    <input type="text" name="object[name]" />
  </label>
</p>
```

A possible solution could be something like:

```ruby
# ...
p do
  label do
    text "Your name?"
    text_field :name
  end
end
```

What do you guys think?

joneslee85:
I like the idea

cc @AlfonsoUceda @jodosha

josemota:
@joneslee85 referred to the possibility for injection to happen. I'll add a failing test to start and work a solution out.

PR #30, opened 2015-08-14
Add support for text inside other tags.

josemota:
This feature allows for more complex structures to be defined. One example is the one of a label with a radio button or a checkbox:

```ruby
builder.label do
  text "Option 1"
  radio_button :option, 1
end
```

### DIFF
--- a/lib/lotus/helpers/html_helper/html_builder.rb
+++ b/lib/lotus/helpers/html_helper/html_builder.rb
@@ -3,6 +3,7 @@ require 'lotus/utils/class_attribute'
 require 'lotus/utils/escape'
 require 'lotus/helpers/html_helper/empty_html_node'
 require 'lotus/helpers/html_helper/html_node'
+require 'lotus/helpers/html_helper/text_node'
 
 module Lotus
   module Helpers
@@ -250,6 +251,32 @@ module Lotus
         #   html.empty_tag(:xr, id: 'foo', 'data-xyz': 'bar') # => <xr id="foo" data-xyz="bar">
         def empty_tag(name, attributes = nil)
           @nodes << EmptyHtmlNode.new(name, attributes)
+          self
+        end
+
+        # Defines a plain string of text. This particularly useful when you
+        # want to build more complex HTML.
+        #
+        # @param content [String] the text to be rendered.
+        #
+        # @return [self]
+        #
+        # @see Lotus::Helpers::HtmlHelper
+        # @see Lotus::Helpers::HtmlHelper::TextNode
+        #
+        # @example
+        #
+        #   html.label do
+        #     text "Option 1"
+        #     radio_button :option, 1
+        #   end
+        #
+        #   # <label>
+        #   #   Option 1
+        #   #   <input type="radio" name="option" value="1" />
+        #   # </label>
+        def text(content)
+          @nodes << TextNode.new(content)
           self
         end
 
--- a/lib/lotus/helpers/html_helper/text_node.rb
+++ b/lib/lotus/helpers/html_helper/text_node.rb
@@ -1,0 +1,30 @@
+module Lotus
+  module Helpers
+    module HtmlHelper
+      # Text node. Allows for text to be inserted between HTML tags.
+      #
+      # @since x.x.x
+      # @api private
+      class TextNode
+        # Initialize a new text node
+        #
+        # @param content [String] The content to be added.
+        #
+        # @return [Lotus::Helpers::HtmlHelper::TextNode]
+        def initialize content
+          @content = content
+        end
+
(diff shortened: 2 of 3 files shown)


josemota:
This refers to #29.

lib/lotus/helpers/html_helper/html_builder.rb
```diff
@@ -253,6 +253,32 @@ def empty_tag(name, attributes = nil)
...
+        #   html.label do
+        #     text "Option 1"
+        #     radio_button :option, 1
+        #   end
+        #
+        #   # <label>
+        #   #   Option 1
+        #   #   <input type="radio" name="option" value="1" />
+        #   # </label>
+
```
joneslee85: nitpick, pls remove this empty line

joneslee85:
please squash to 1 commit, it is :+1:  from me

lib/lotus/helpers/html_helper/text_node.rb
```diff
@@ -0,0 +1,30 @@
+module Lotus
+  module Helpers
+    module HtmlHelper
+      # Text node. Allows for text to be inserted in between HTML tags.
+      #
+      # @since 0.1.0
```
AlfonsoUceda: use `x.x.x` please

lib/lotus/helpers/html_helper/text_node.rb
```diff
@@ -0,0 +1,30 @@
+module Lotus
+  module Helpers
+    module HtmlHelper
+      # Text node. Allows for text to be inserted in between HTML tags.
```
AlfonsoUceda: this is `inserted in between` or `inserted between` ?

lib/lotus/helpers/html_helper/text_node.rb
```diff
@@ -0,0 +1,30 @@
...
+        # @return [Lotus::Helpers::HtmlHelper::TextNode]
+        def initialize content
+          @content = content
+        end
+
+        # Resolve and return the output
+        #
+        # @return [String] the output
+        #
+        # @since 0.1.0
```
AlfonsoUceda: change to `x.x.x`

test/html_helper/html_builder_test.rb
```diff
@@ -196,4 +196,30 @@
...
+
+      result.must_equal(%(<p>\n<span>Foo</span>\nBar\n</p>))
+    end
+
+    it "escapes HTML inside" do
+      result = @builder.text(%(<p>Foo</p>)).to_s
+      result.must_equal('&lt;p&gt;Foo&lt;&#x2F;p&gt;')
+    end
+  end
+
```
AlfonsoUceda: remove this extra line

AlfonsoUceda:
@josemota :ok: after feedback ;) and remember squash to one or use push force for only one commit

thank you very much

josemota:
Got it, thanks @AlfonsoUceda ! Will try and fix the changes today.

jodosha:
Hi @josemota great addition and implementation. :clap: 
I've added another way to do concatenation, via `+` operator (see https://github.com/lotus/helpers/commit/6671aa119d970cc7ce99ede44d8d065458da33e3):

```ruby
builder.label do
  span('Foo') +
    'Bar'
end
```

@lotus/core-team @josemota Do you thing that `text` is a good name for this purpose? My concern is that, W3C or browser vendors may introduce a `<text>` tag in the future. That would mean to rename ours with a different name.

josemota:
Hey @jodosha ! Thanks for the feedback. Great addition by the way, `+` is awesome. :+1: 

I thought of it as the best way to represent what it is to be... just text :P If the possibility for the `<text>` tag is real then yea, we should think of another name for the method. `content` comes to mind but I'm not sure. Suggestions?

jodosha:
@josemota It's hard to foretell what the future deserves :space_invader: There is a growing list of tags (see [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)), and `<content>` is [already there](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/content) :wink: 

Let's hope for the best. :four_leaf_clover:

pascalbetz:
what about _plain_ ? or _plain_text_?

josemota:
I thought of that too but didn't think anyone would too :P so :+1: for me.
On Sep 8, 2015 11:30, Pascal Betz <notifications@github.com> wrote:what about plain ? or plain_text?

AlfonsoUceda:
and `no_tag`? :P

josemota:
http://laughingmeme.org/2005/12/23/there-are-only-two-hard-things-in-computer-science-cache-invalidation-and-naming-things/

Issue #32, opened 2015-08-21
`text_area` method is not providing either `id` or `name` attributes


josemota:
The title is self-explanatory but here's my own example:

**Code**

```ruby
form_for :event, "/events", class: "row" do
  text_area :notes, placeholder: "Some text here.", class: "form-control", rows: 6
end
```

**Output**

```html
<form action="/events" method="post" class="row" id="event-form">
  <textarea rows="6" class="form-control" placeholder="Some text here"></textarea>
</form>
```

It is expected that the textarea has:

* A `name` attribute with the value of `event[notes]`;
* An `id` attribute with the value of `event-note`.

* * *

EDIT: The root cause of the bug is the inability to store both name and id when no content is passed in. Check [the discourse](https://github.com/lotus/helpers/issues/32#issuecomment-135385365)

joneslee85:
Good find, would you be able to submit a PR? TIA

pascalbetz:
Without looking into the code, shouldn't it be:

```ruby
form_for :event, "/events", class: "row" do |f|
  f.text_area :notes, placeholder: "Some text here.", class: "form-control", rows: 6
end
```
(Form builder passed as a block argument).

joneslee85:
@pascalbetz nope

eg:
```
form_for :book, routes.books_path do
    text_field :title

    submit 'Create'
  end
```

josemota:
@joneslee85 yes, absolutely. You'll need to give me some time, though. My wife is in the hospital, it seems that my baby wants to be born any hour now. :) I'll be in touch.

AlfonsoUceda:
@joneslee85 you are right, it should be like the example https://github.com/lotus/helpers#view-usage

pascalbetz:
@joneslee85  i was referring to the first example at http://lotusrb.org/guides/helpers/forms/ but without checking the text below. Argh.

@josemota : all the best!

josemota:
@joneslee85 @pascalbetz My baby's been born! She's growing steadily now in the incubator for some time. Until then, I'll be on the lookout but I'll try and squeeze in some time to work something out.

Please bear with me :)

josemota:
I've found the cause of this bug. It seems that not passing any content to the method will simply not register either `name` or `id`.

```ruby
text_area :notes, "forced content", rows: 1 # expected behavior is met
text_area :notes, rows: 1 # expected behavior is NOT met
```

[...]

PR #33, opened 2015-08-27
Fix attribution of HTML attributes.

josemota:
Relates to #32.
    
In the scenario of having extra HTML attributes without some content being
passed in, the attributes hash was being set to nil while the content variable
holding the attributes hash.
    
In the scenario:
    
```ruby
form_for :book do
  text_area :description, rows: 5
end
```

The `content` argument was `{ rows: 5 }` and `attributes` was `{}`.

This is still a pretty ugly solution. I need some feedback on this.

### DIFF
--- a/lib/lotus/helpers/form_helper/form_builder.rb
+++ b/lib/lotus/helpers/form_helper/form_builder.rb
@@ -526,6 +526,11 @@ module Lotus
         #   # Output:
         #   #  <textarea name="user[hobby]" id="user-hobby">Football</textarea>
         def text_area(name, content = nil, attributes = {})
+          if content.is_a? Hash
+            attributes = content
+            content = nil
+          end
+
           attributes = {name: _input_name(name), id: _input_id(name)}.merge(attributes)
           _content = content || _value(name)
           textarea(_content, attributes)
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -762,6 +762,14 @@ describe Lotus::Helpers::FormHelper do
       actual.must_include %(<textarea name="book[description]" id="book-description" class="form-control" cols="5"></textarea>)
     end
 
+    it "renders all HTML attributes along with the name and id" do
+      actual = view.form_for(:book, action) do
+        text_area :description, class: 'form-control', cols: '5'
+      end.to_s
+
+      actual.must_include %(<textarea name="book[description]" id="book-description" class="form-control" cols="5"></textarea>)
+    end
+
     describe "set content explicitly" do
       let(:content) { "A short description of the book" }
       it "allows to set content" do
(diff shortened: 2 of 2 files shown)


joneslee85:
@josemota thanks for the PR and congrats on your newborn

On the second look, I dislike the API interface of `text_area`, IMHO I think the content of text area should be parased as block, if no block given, default to blank. Thoughts? @jodosha @AlfonsoUceda

pascalbetz:
i don't like the "can be a hash or a string. if it is a hash, then ..." kind of interfaces.

text_area is a special case because its content is specified as body not as attribute in HTML, in that it is similar to select and so they should have the same interface. 

Or something like this?
```ruby
def text_area(name, options = {})
  _content = options[:value] # or :content or ...
  attributes = {name: _input_name(name), id: _input_id(name)}.merge(attributes)
  textarea(_content, attributes)
end
```

But i'm not sure how the select helper would look like.

AlfonsoUceda:
@josemota have you tried in the merge this? `merge(attributes || content)` ?

josemota:
My review on your thoughtful feedback:

> i don't like the "can be a hash or a string. if it is a hash, then ..." kind of interfaces.

@pascalbetz i agree, it's awful. Still, changing the API signature breaks the tests.

@AlfonsoUceda I'll try and explore that idea. Sounds way cleaner than my ugly ass `if`. :hankey:

pascalbetz:
@josemota / @joneslee85 i find a block somewhat strange because the value (a.k.a. body) of a _text_field_ is always a string. I could understand for _select_ since there it's a complex object (options, possible opt groups). I think the API should be the same across all form helpers.

Breaking an API is no problem to me. It's required to improve software.

joneslee85:
@pascalbetz I don't want to break API interface right now, let's bring this discussion into discuss.lotusrb.org. Can you please help create one topic?

For the time being, I am happy with whatever it is

AlfonsoUceda:
@joneslee85 @pascalbetz I think my approach doesn't break API, @joneslee85 can we revert the merge and wait if @josemota can try my suggestion?

josemota:
Even though I appreciate your approval, it's kinda of an ugly solution and I don't particularly like it. I'm still digging into a better solution but I'll keep this thread posted. Upon new developments and a possible new PR for refactoring, I'll notify you guys.

AlfonsoUceda:
@josemota ok thanks, make a new PR when you feel comfortable ;)

jodosha:
@josemota Thanks for this fix. :+1:  I added some extra tests at https://github.com/lotus/helpers/commit/bb1694ba5905cf62b58d475c9840bae1d1c5973a

josemota:
:+1:

PR #40, opened 2015-10-29
Renaming method to verb_method on FormBuilder

rpanachi:
Applying feedback from this https://github.com/lotus/helpers/pull/39

@joneslee85 @jodosha for now, I changed to `verb_method`, because I don't figured out why the `@verb` variable was defined only on `_method_override!`. If it was defined before the `to_s` method, a lot os tests was broken... Maybe this was a smell to refactoring how the `@verb` variable should be handled.

Thanks


jodosha:
@rpanachi Wonderful :+1:

PR #41, opened 2015-11-24
Add video tag helper

leighhalliday:
https://github.com/lotus/lotus/issues/396

This commit exposes a `video` method to the view
which can be used to generate HTML5 video tags.
It accepts options to generate HTML attributes along
with a block for creating fallback content and/or
`source` and `track` tags.

### DIFF
--- a/lib/lotus/helpers/video_helper.rb
+++ b/lib/lotus/helpers/video_helper.rb
@@ -1,0 +1,86 @@
+module Lotus
+  module Helpers
+    module VideoHelper
+      include Lotus::Helpers::HtmlHelper
+
+      # Generates a video tag for the given arguments.
+      #
+      # Contents are automatically escaped.
+      #
+      # @overload link_to(content, url, options)
+      #   Use string as content
+      #   @param content [String] content used in the a tag
+      #   @param url [String] url used in href attribute
+      #   @param options [Hash] HTML attributes to pass to the a tag
+      #
+      # @overload link_to(url, options, &blk)
+      #   Use block as content
+      #   @param url [String] url used in href attribute
+      #   @param options [Hash] HTML attributes to pass to the a tag
+      #   @param blk [Proc] A block that describes the contents of the a tag
+      #
+      # @return [String] HTML markup for the video
+      #
+      # @raise [ArgumentError] if the signature isn't respected
+      #
+      # @since 0.5.0
+      #
+      # @see Lotus::Helpers::HtmlHelper#html
+      #
+      # @example Basic usage
+      #   <%= video('movie.mp4') %>
+      #     # => <video src="/assets/movie.mp4"></video>
+      #
+      # @example HTML attributes
+      #   <%= video('movie.mp4', autoplay: true, controls: true) %>
+      #     # => <video src="/assets/movie.mp4" autoplay="autoplay" controls="controls"></video>
+      #
+      # @example Fallback Content
+      #   <%=
+      #     video('movie.mp4') do
+      #       "Your browser does not support the video tag"
+      #     end
+      #   %>
+      #     # => <video src="/assets/movie.mp4">\nYour browser does not support the video tag\n</video>
+      #
+      # @example Tracks
+      #   <%=
+      #     video('movie.mp4') do
+      #       track kind: 'captions', src: view.asset_path('movie.en.vtt'), srclang: 'en', label: 'English'
+      #     end
+      #   %>
+      #     # => <video src="/assets/movie.mp4">\n<track kind="captions" src="/assets/movie.en.vtt" srclang="en" label="English">\n</video>
+      #
+      # @example Sources
+      #   <%=
+      #     video do
+      #       text "Your browser does not support the video tag"
+      #       source src: view.asset_path('movie.mp4'), type: 'video/mp4'
+      #       source src: view.asset_path('movie.ogg'), type: 'video/ogg'
+      #     end
+      #   %>
+      #     # => <video>\nYour browser does not support the video tag\n<source src="/assets/movie.mp4" type="video/mp4">\n<source src="/assets/movie.ogg" type="video/ogg">\n</video>
+      #
+      # @example Without any argument
+      #   <%= video %>
+      #     # => ArgumentError
+      #
+      # @example Without src and without block
+      #   <%= video(content: true) %>
+      #     # => ArgumentError
+      def video(src = nil, options = {}, &blk)
+        if src.is_a?(String)
+          options = {src: asset_path(src)}.merge(options)
+        elsif src.is_a?(Hash)
+          options = src
+        end
+
+        if !options[:src] && !block_given?
+          raise ArgumentError
+        end
+
+        html.video(blk, options).to_s
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


leighhalliday:
For my test view I created an `asset_path` method because it hasn't yet been completed. I took the assumption that this will be the method doing the work of choosing how exactly to generate the digest URL or prefixed URL.

test/video_helper_test.rb
```diff
@@ -0,0 +1,53 @@
+require 'test_helper'
+
+describe Lotus::Helpers::VideoHelper do
+  let(:view)   { VideoHelperView.new(params) }
+  let(:params) { Hash[] }
+
+  #
+  # VIDEO
```
joneslee85: I think the comment is not needed

joneslee85:
looks good :+1:

test/video_helper_test.rb
leighhalliday: Thanks Trung, I've removed the comment :)

joneslee85:
cc @lotus/core-team

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+      #   Use block as content
+      #   @param url [String] url used in href attribute
+      #   @param options [Hash] HTML attributes to pass to the a tag
+      #   @param blk [Proc] A block that describes the contents of the a tag
+      #
+      # @return [String] HTML markup for the video
+      #
+      # @raise [ArgumentError] if the signature isn't respected
+      #
+      # @since 0.5.0
```
jodosha: `@since x.x.x`

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+      # @example Without any argument
+      #   <%= video %>
+      #     # => ArgumentError
+      #
+      # @example Without src and without block
+      #   <%= video(content: true) %>
+      #     # => ArgumentError
+      def video(src = nil, options = {}, &blk)
+        if src.is_a?(String)
+          options = {src: asset_path(src)}.merge(options)
```
jodosha: @leighhalliday Where is this `asset_path` defined?

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+        elsif src.is_a?(Hash)
+          options = src
+        end
+
+        if !options[:src] && !block_given?
+          raise ArgumentError
```
jodosha: @leighhalliday I think we should be more specific on this. What if we add a message like "You should provide a source via `src' option or with a `source' HTML tag"?

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
+module Lotus
+  module Helpers
+    module VideoHelper
+      include Lotus::Helpers::HtmlHelper
+
+      # Generates a video tag for the given arguments.
+      #
+      # Contents are automatically escaped.
+      #
+      # @overload link_to(content, url, options)
```
AlfonsoUceda: Should these overload sections be removed?

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+          options = {src: asset_path(src)}.merge(options)
+        elsif src.is_a?(Hash)
+          options = src
+        end
+
+        if !options[:src] && !block_given?
+          raise ArgumentError
+        end
+
+        html.video(blk, options).to_s
```
jodosha: @leighhalliday We shouldn't call `#to_s` prematurely here. It will be invoked by a ERb. It's precious to have a more complex returning value here, for things like method chaining.

test/video_helper_test.rb
```diff
@@ -0,0 +1,49 @@
+require 'test_helper'
```
jodosha: This is great, thank you! :+1: 

Can you please also add an integration test with Lotus::View? See https://github.com/lotus/helpers/tree/master/test/integration

It's important, because these helpers will be used in combination with views that perform automatic escape of contents. We must render a real template and assert that the generated HTML is correct. :wink:

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+      #     # => <video>\nYour browser does not support the video tag\n<source src="/assets/movie.mp4" type="video/mp4">\n<source src="/assets/movie.ogg" type="video/ogg">\n</video>
+      #
+      # @example Without any argument
+      #   <%= video %>
+      #     # => ArgumentError
+      #
+      # @example Without src and without block
+      #   <%= video(content: true) %>
+      #     # => ArgumentError
+      def video(src = nil, options = {}, &blk)
```
jodosha: @leighhalliday If options is explicitly set as `nil` this method will raise a `TypeError` :wink:

lib/lotus/helpers/video_helper.rb
```diff
@@ -0,0 +1,86 @@
...
+        if src.is_a?(String)
```
jodosha: @leighhalliday For compatibility with other libs it's counterproductive to use `is_a?`. Let's say some other lib has an object called `Options` which can be serialized as a Hash. With this implementation we'll totally ignore that type. This doesn't make our code "Open for extension, Closed for modifications".

As a side note:

```ruby
[1] pry(main)> require 'lotus/utils/hash'
=> true
[2] pry(main)> Lotus::Utils::Hash.new.is_a?(Hash)
=> false
```

This code isn't compatible with Lotus itself :wink: 

What if we do:

```ruby
if src.respond_to?(:to_hash)
  options = src.to_hash
else
  options[:src] = asset_path(src)
end
```

That will make the implementation compatible with `Hash`, `Lotus::Utils::Hash`, any eventual object that can serialize itself into a Hash.

In the `else` condition, we'll leave delegate `asset_path` to coerce into a string.

We don't use `merge` here, which allocates a new Hash that is not really needed. See https://speakerdeck.com/sferik/writing-fast-ruby :wink:

jodosha:
@leighhalliday Hey thanks for taking care of this. I left a few comments in the code. But I've got a more wider question to ask:

Why did you opened a pull request here instead of https://github.com/lotus/lotus ?
The original ticket was there :wink: 

Now, I think that this is a feature that belongs to the `lotusrb` "glue code realm". It's an intersection of `lotus-helpers` because of the HTML helpers, of `lotus-assets` because this is related to static assets and because it will depend on `asset_path`, which is tight to `Lotus::Assets` concepts such as "Digest Mode" (aka checksum prefix in asset filename), "Path Prefix" (eg. `/assets/admin/foo.js`) or to the knowledge of scheme/host/port.

In `lotusrb` we already inject "intersectional" modules. For instance `Lotus::Action::RoutingHelpers` isn't defined by `lotus-controller` but `lotusrb`, because that needs to know about the implementation of an action, and how Lotus creates factories for routes.

**Can you please implement this feature in `lotusrb`? Thank you very much!**

leighhalliday:
Thanks for your comments @jodosha!! I'll try to address the changes in the next couple days. Thanks for taking the time to give me such an in depth PR review :)

PR #42, opened 2015-12-04
Add definition list element 'dd'

skyriser:
I noticed HtmlBuilder not have definition list element 'dd'.

http://www.w3.org/TR/html4/struct/lists.html#h-10.3

I think it is required.


AlfonsoUceda:
:+1:  cc @lotus/core-team

joneslee85:
Good to me

skyriser:
thx!

PR #44, opened 2015-12-31
Propose html fragment

huynhquancam:
Hello @lotus,

I would like to propose html fragment (or maybe another better name) which will be useful to avoid html string concatenation like

```ruby
html.div("Hello") + html.div("Lotus") + html.div("Ruby")
#=> <div>Hello</div><div>Lotus</div><div>Ruby</div>
```

It looks much better to me with html fragment declaration, especially nested HTML.

```ruby
html do
  div "Hello"
  div "Lotus"
  div "Ruby"
end
```

### DIFF
--- a/lib/lotus/helpers/html_helper.rb
+++ b/lib/lotus/helpers/html_helper.rb
@@ -65,6 +65,15 @@ module Lotus
     #   #  hello
     #   #</div>
     #
+    #   # 8
+    #   html do
+    #     li 'Hello'
+    #     li 'Lotus'
+    #   end
+    #   # =>
+    #   #<li>Hello</li>
+    #   #<li>Lotus</li>
+    #
     #
     #
     # @example Complex markup
@@ -178,14 +187,20 @@ module Lotus
       private
       # Instantiate an HTML builder
       #
+      # @param blk [Proc,Lotus::Helpers::HtmlHelper::HtmlBuilder,NilClass] the optional content block
+      #
       # @return [Lotus::Helpers::HtmlHelper::HtmlBuilder] the HTML builder
       #
       # @since 0.1.0
       #
       # @see Lotus::Helpers::HtmlHelper
       # @see Lotus::Helpers::HtmlHelper::HtmlBuilder
-      def html
-        HtmlBuilder.new
+      def html(&blk)
+        if block_given?
+          HtmlBuilder.new.fragment(&blk)
+        else
+          HtmlBuilder.new
+        end
       end
     end
   end
--- a/lib/lotus/helpers/html_helper/html_builder.rb
+++ b/lib/lotus/helpers/html_helper/html_builder.rb
@@ -3,6 +3,7 @@ require 'lotus/utils/class_attribute'
 require 'lotus/utils/escape'
 require 'lotus/helpers/html_helper/empty_html_node'
 require 'lotus/helpers/html_helper/html_node'
+require 'lotus/helpers/html_helper/html_fragment'
 require 'lotus/helpers/html_helper/text_node'
 
 module Lotus
@@ -229,6 +230,33 @@ module Lotus
         #   #</custom>
         def tag(name, content = nil, attributes = nil, &blk)
           @nodes << HtmlNode.new(name, blk || content, attributes || content, options)
+          self
+        end
+
+        # Define a HTML fragment
+        #
+        # @param content [String,Lotus::Helpers::HtmlHelper::HtmlBuilder,NilClass] the optional content
+        # @param blk [Proc] the optional nested content espressed as a block
+        #
+        # @return [self]
+        #
+        # @since 0.2.6
+        # @api public
+        #
+        # @see Lotus::Helpers::HtmlHelper
+        #
+        # @example
+        #   html.fragment('Lotus') # => Lotus
+        #
+        #   html do
+        #     p 'hello'
+        #     p 'lotus'
+        #   end
+        #   # =>
+        #     <p>hello</p>
+        #     <p>lotus</p>
+        def fragment(&blk)
+          @nodes << HtmlFragment.new(&blk)
           self
         end
 
--- a/lib/lotus/helpers/html_helper/html_fragment.rb
+++ b/lib/lotus/helpers/html_helper/html_fragment.rb
@@ -1,0 +1,45 @@
+module Lotus
+  module Helpers
+    module HtmlHelper
+      # HTML Fragment
+      #
+      # @since 0.2.6
+      # @api private
+      #
+      # @see Lotus::Helpers::HtmlHelper::HtmlFragment
+      class HtmlFragment
+        # Initialize a HTML Fragment
+        #
+        # @param blk [Proc,Lotus::Helpers::HtmlHelper::HtmlBuilder,NilClass] the content block
+        #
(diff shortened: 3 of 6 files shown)


joneslee85:
@huynhquancam I was caught by surprise at first, it is :+1: to me though I am not very sold with the naming `Fragment`.

cc @lotus/core-team

huynhquancam:
@joneslee85 I am not very happy with `Fragment` too, just randomly grabbed the concept that from `Nokogiri`.

How about `Context` as advised, `Block`, or `Collection`?

lib/lotus/helpers/html_helper.rb
```diff
@@ -176,7 +185,7 @@ module Helpers
     #     #    </div>
     module HtmlHelper
       private
-      # Instantiate an HTML builder
+      # Instantiate an HTML builder / HTML Fragment
```
jodosha: This fragment name is an internal detail that we shouldn't document here.
For developers it's just a method that let them to build HTML.

Can you please remove it?

lib/lotus/helpers/html_helper.rb
```diff
@@ -184,8 +193,13 @@ module HtmlHelper
       #
       # @see Lotus::Helpers::HtmlHelper
       # @see Lotus::Helpers::HtmlHelper::HtmlBuilder
-      def html
-        HtmlBuilder.new
+      # @see Lotus::Helpers::HtmlHelper::HtmlFragment
+      def html(&blk)
+        if block_given?
+          HtmlBuilder.new.fragment(&blk)
```
jodosha: @huynhquancam Why not just simplify this call with the following syntax?

```ruby
def html(&blk)
  HtmlBuilder.new(&blk)
end
```

And then in `HtmlBuilder`:

```ruby
def initialize(&blk)
  # ...
  instance_eval(&blk) if block_given?
end
```

lib/lotus/helpers/html_helper/html_builder.rb
```diff
@@ -232,6 +233,33 @@ def tag(name, content = nil, attributes = nil, &blk)
...
+        #
+        #   html do
+        #     p 'hello'
+        #     p 'lotus'
+        #   end
+        #   # =>
+        #     <p>hello</p>
+        #     <p>lotus</p>
+        def fragment(content = nil, &blk)
+          @nodes << HtmlFragment.new(blk || content)
```
jodosha: @huynhquancam Why this fragment concept is needed at all? If we pass a block to `#initialize` we can `instance_eval` it.

I don't like the idea of `html.fragment('Lotus')`, it exposes an internal concept. The expected API is a mapping for HTML tags. For instance, to expose `#div` and `#span` for `<div>` and `<span>`, respectively.

If we still need this method and `HtmlFragment`, it should only accept `&blk` and not `content` as well.

lib/lotus/helpers/html_helper.rb
```diff
@@ -184,8 +193,13 @@ module HtmlHelper
       #
```
jodosha: @huynhquancam Please document `&blk`.

test/html_helper/html_builder_test.rb
```diff
@@ -146,6 +146,27 @@
...
+    it 'generates a html fragment' do
+      result = @builder.fragment do
+        span 'Hello'
+        span 'Lotus'
+      end.to_s
+
+      result.must_equal "\n<span>Hello</span>\n<span>Lotus</span>\n"
+    end
+
+    it 'generates a html fragment with content' do
```
jodosha: @huynhquancam As mentioned above, this behavior should be removed.

jodosha:
@huynhquancam I like the idea, thank you, it's worth to be included. :+1: 
I left some notes, to improve the implementation. Can you please rework it?

Can you please also add an two extra tests to verify that two `html` blocks can be still concat?

1.

```ruby
html.div('Hello') + html.div('World')
```

2.

```ruby
html do
  div "Hello"
end +
  html do
    div "World"
  end
```

Thank you! :confetti_ball:

lib/lotus/helpers/html_helper.rb
huynhquancam: @jodosha thanks, do you think if it's viable to make it work this way?

```ruby
# html_builder.rb
def initialize(&blk)
  @nodes = []
  if block_given?
    @nodes = instance_eval(&blk).nodes
  end
end

attr_reader :nodes
```

Also `FormBuilder` raises conflict when we allow block as a parameter in `HtmlBuilder#initialize`. It could be bypassed by removing the `super()` invocation, but I can foresee one of disadvantages is about the reusability of `HTMLBuilder`.

```ruby
class FormBuilder
  def initialize(form, attributes, context = nil, &blk)
    @nodes = []
    # instead of super() invocation
    # ...
  end
end
```

jodosha: @huynhquancam Good point about `FormBuilder` :+1: Never mind, please leave the initializer as it is.

Only one thing about `#html`. For performance reasons is it possible to avoid to bind `&blk` for each call? In other words:

```ruby
def html
  builder = HtmlBuilder.new
  builder.fragment { yield } if block_given?
  builder
end
```

huynhquancam: @jodosha I'm afraid not, as `yield` is supposed to use the context where the block was defined.

I actually tried to executing this in the test with `yield`

```ruby
html do
  div 'Lotus'
end

# => NoMethodError: undefined method `div' for #<HtmlView:0x007fe5422fdaa8>
```

jodosha: @huynhquancam Never mind again :sweat_smile: 

Let's the production code as it is, but please add the other tests and fix the docs, then it's mergeable! Thank you very much! :+1:

huynhquancam:
@jodosha @joneslee85 Thanks for your reviews :+1:. I've updated the code as requested. :smile:.

Kindly let me know if I should squash my commits. Thanks.

joneslee85:
@huynhquancam please squash

lib/lotus/helpers/html_helper/html_builder.rb
```diff
@@ -232,6 +233,33 @@ def tag(name, content = nil, attributes = nil, &blk)
...
         end
 
+        # Define a HTML fragment
+        #
+        # @param content [String,Lotus::Helpers::HtmlHelper::HtmlBuilder,NilClass] the optional content
+        # @param blk [Proc] the optional nested content espressed as a block
+        #
+        # @return [self]
+        #
+        # @since 0.1.0
```
joneslee85: is it since 0.1.0?!

lib/lotus/helpers/html_helper/html_fragment.rb
```diff
@@ -0,0 +1,45 @@
+module Lotus
+  module Helpers
+    module HtmlHelper
+      # HTML node
+      #
+      # @since 0.2.6
```
joneslee85: please fix the version

huynhquancam:
@joneslee85 thanks, commits squashed and version updated :)

joneslee85:
:+1:

jodosha:
@huynhquancam I love it! Thank you very much!

huynhquancam:
@jodosha awesome, thanks for your help!

Issue #48, opened 2016-01-25
Errors with calling link_to helper in html builder


davydovanton:
When I call `link_to` helper in `html` helper I have a problems.
I don't know solution for this problem, but [I have a patch with test cases](https://github.com/davydovanton/helpers/commit/814e8b3d8e552fcb525bdd09eff9244669712c6f):

``` diff
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -569,3 +569,22 @@ def link_to_with_content_html_content_id_and_class
   end
 
 end
+
+class HtmlAndLinkTo
+  include Hanami::Helpers::HtmlHelper
+  include Hanami::Helpers::LinkToHelper
+
+  def two_links_to_in_div
+    html.div do
+      link_to('Posts', 'posts')
+      link_to('Comments', 'comments')
+    end
+  end
+
+  def span_and_link_to_in_div
+    html.div do
+      span 'hello'
+      link_to('Comments', 'comments')
+    end
+  end
+end
diff --git a/test/html_helper_test.rb b/test/html_helper_test.rb
index 1fd6d6f..6487a43 100644
--- a/test/html_helper_test.rb
+++ b/test/html_helper_test.rb
@@ -92,4 +92,18 @@
   it 'autoescapes nested blocks' do
     @view.evil_nested_block_content.to_s.must_equal %(<div>\n<p>&lt;script&gt;alert(&apos;xss&apos;)&lt;&#x2F;script&gt;</p>\n</div>)
   end
+
+  describe 'with link_to helper' do
+    before do
+      @view = HtmlAndLinkTo.new
+    end
+
+    it 'returns two links in div' do
+      @view.two_links_to_in_div.to_s.must_equal %(<div>\n<a href=\"comments\">Comments</a>\n<a href=\"posts\">Posts</a>\n</div>)
+    end
+
+    it 'returns span and link in div' do
+      @view.span_and_link_to_in_div.to_s.must_equal %(<div>\n<span>hello</span>\n<a href=\"posts\">Posts</a>\n</div>)
+    end
+  end
 end
```

davydovanton:
May be is better to create alias from `link_to` to `a` tag? What do you think?

joneslee85:
@davydovanton what was the output of the error?

davydovanton:
@joneslee85, sorry I forgot to add this :disappointed: 

```

  1) Failure:
[...]

PR #52, opened 2016-02-04
Allow links to be concat

jodosha:
Ref #48

### DIFF
--- a/lib/hanami/helpers/link_to_helper.rb
+++ b/lib/hanami/helpers/link_to_helper.rb
@@ -128,7 +128,7 @@ module Hanami
           raise ArgumentError
         end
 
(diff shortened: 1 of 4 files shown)


davydovanton:
nice :+1:

But also you need to realize `#html` helper too. Because you will have a problems with following code now too.

``` ruby
def two_spans_in_div
  html.div do
    span 'hello'
    html.span 'world'
  end
end
```

jodosha:
@davydovanton What's the point of using nested `html` calls?

davydovanton:
My hypothesis is that perhaps we'll have a problems with helpers based on `html` helper inside html block.
But now I see that all tests is passed.

PR #53, opened 2016-02-14
Add information about multiple option

davydovanton:


joneslee85:
:+1:

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -486,6 +486,7 @@ def hidden_field(name, attributes = {})
         # @param name [Symbol] the input name
         # @param attributes [Hash] HTML attributes to pass to the input tag
         # @option attributes [String,Array] :accept Optional set of accepted MIME Types
+        # @option attributes [TrueClass,FalseClass] :multiple Optional, allow multiple file upload
```
joneslee85: what would be the output if user provides `false`?

davydovanton: this method will return `<input>` tag without `multiple` option

joneslee85: is there test for this option?

davydovanton: [yep](https://github.com/hanami/helpers/blob/master/test/form_helper_test.rb#L661-L667)

davydovanton:
thanks :+1:

jodosha:
@davydovanton @joneslee85 Thank you :+1:

PR #54, opened 2016-02-20
Support include_blank option for select form helper

davydovanton:


joneslee85:
@davydovanton have you ever thought about the default option could have different messages? In Rails 4, there is `prompt` option.

AlfonsoUceda:
@davydovanton thanks for your PR!

I prefer `prompt` option because a person could add something like `Select option` to the first option and we should allow an empty string like `include_blank`does.

davydovanton:
@joneslee85 @AlfonsoUceda thanks for feedback :+1: 
Yes, I think you're right. I'll update this PR soon, thanks!

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -751,11 +751,27 @@ def password_field(name, attributes = {})
         #   #    <option value="it" selected="selected">Italy</option>
         #   #    <option value="us">United States</option>
         #   #  </select>
+        #
+        # @example Include blank option
+        #   <%=
+        #     # ...
+        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     select :stores, values, options: { include_blank: true}
```
joneslee85: nitpick, 

```
{ include_blank: true}
#to
{include_blank: true}
```

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -751,11 +751,27 @@ def password_field(name, attributes = {})
...
+        #   %>
+        #
+        #   # Output:
+        #   #  <select name="book[store]" id="book-store">
+        #   #    <option></option>
+        #   #    <option value="it">Italy</option>
+        #   #    <option value="us">United States</option>
+        #   #  </select>
         def select(name, values, attributes = {})
           options    = attributes.delete(:options) || {}
```
joneslee85: what do you think about parsing the block in the case that key is not found?

```
options = attributes.delete(:options) { {} )
```

jodosha:
@davydovanton PING :wink:

lib/hanami/helpers/form_helper/form_builder.rb
davydovanton: wow, I don't know about this option, thanks! :+1:

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -751,11 +751,42 @@ def password_field(name, attributes = {})
...
+        #   #  </select>
         def select(name, values, attributes = {})
-          options    = attributes.delete(:options) || {}
-          attributes = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
+          options      = attributes.delete(:options) { {} }
+          attributes   = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
+          prompt_value = options.delete(:include_blank) || options.delete(:prompt)
 
           super(attributes) do
+            option(prompt_value.is_a?(String) ? prompt_value : '') if prompt_value
```
davydovanton: I don't like this solution, but when `prompt_value` is `true` `prompt_value.to_s` returns `'true'` string. If you have better solution I'll happy to use it :+1:

davydovanton:
@jodosha pong :)
I added `prompt` option and fixed all considerations.

lib/hanami/helpers/form_helper/form_builder.rb
jodosha: @davydovanton I think that `prompt:` should be used only with strings. For instance

```erb
<%=
  values = Hash['it' => 'Italy', 'us' => 'United States']
  select :stores, values, options: {prompt: 'Select a City'}
%>
```

or

```erb
<%=
  values = Hash['it' => 'Italy', 'us' => 'United States']
  select :stores, values, options: {prompt: ''}
%>
```

but not..

```erb
<%=
  values = Hash['it' => 'Italy', 'us' => 'United States']
  select :stores, values, options: {prompt: true}
%>
```

Does it makes sense?

davydovanton: hmmm, I think you're right. thanks!

davydovanton: done.

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -751,11 +751,44 @@ def password_field(name, attributes = {})
...
+        #   #    <option>Select a City</option>
+        #   #    <option value="it">Italy</option>
+        #   #    <option value="us">United States</option>
+        #   #  </select>
         def select(name, values, attributes = {})
-          options    = attributes.delete(:options) || {}
-          attributes = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
+          options      = attributes.delete(:options) { {} }
+          attributes   = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
+          blank_value  = options.delete(:include_blank)
```
jodosha: @davydovanton `include_blank: true` overlaps with `prompt: ""`. What if we remove the first? :wink:

jodosha:
@davydovanton I've just merged this, by removing the `include_blank` and allowing only `prompt`. Thanks :+1:

davydovanton:
thanks!

PR #60, opened 2016-03-05
adding datalist form helper

### LINKED ISSUE #57: implementation of datalist helper
Hi,

I'd like to open this ticket to ask for implementation of the datalist helper: https://github.com/hanami/helpers/blob/23a7df457513ad79bd1d3da5a4c8f1cea5776923/lib/hanami/helpers/html_helper/html_builder.rb#L42 .

Thank you.
Kind regards,
Sebastjan

hoksilato:
Adding `datalist` form helper, fix #57 

- [x] Basic render

    ```html
   <input list="browsers" name="browser">
  <datalist id="browsers">
    <option value="Internet Explorer">
    <option value="Firefox">
    <option value="Chrome">
    <option value="Opera">
    <option value="Safari">
  </datalist>
    ```
- [x] Just allows to override 'id' attribute of the text input
- [x] Allows to override 'name' attribute of the text input
- [x] Allows to specify HTML attributes of the text input
- [x] Allows to specify HTML attributes for datalist
- [x] Allows to specify HTML attributes for options

The only problem I think we need to fix is about `option` tag, it's `CONTENT_TAGS` on `html_builder.rb` so that it will have close tag.

This is my first pull request for Ruby/Hanami (I just learned Ruby about 3 months ago and didn't work much with it). Please let me know if there's anything wrong.

Thank you.

P/S: sorry for my bad English.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -766,6 +766,45 @@ module Hanami
           end
         end
 
+        # Datalist input
+        #
+        # @param name [Symbol] the input name
+        # @param values [Hash] a Hash to generate <tt><option></tt> tags of datalist.
+        #   Keys correspond to <tt>value</tt> and values correspond to the content.
+        # @param list [String] the name of list for the text input, it's also the id of datalist
+        # @param attributes [Hash] HTML attributes to pass to the input tag
+        #
+        # @since 0.3.1
+        #
+        # @example Basic usage
+        #   <%=
+        #     # ...
+        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     datalist :stores, values, 'books'
+        #   %>
+        #
+        #   # Output:
+        #   #  <input type="text" name="book[store]" id="book-store" list="books">
+        #   #  <datalist id="books">
+        #   #    <option value="Italy"></option>
+        #   #    <option value="United States"></option>
+        #   #  </datalist>
+        def datalist(name, values, list, attributes = {})
+          options  = attributes.delete(:options) || {}
+          datalist = attributes.delete(:datalist) || {}
+          attributes = { type: 'text', name: _input_name(name), id: _input_id(name), list: list }.merge(attributes)
+
+          # remove 'id' attribute
+          datalist = { id: list }.merge(datalist)
+
+          input(attributes)
+          super(datalist) do
+            values.each do |value, content|
+              option(value, {value: content}.merge(options))
+            end
+          end
+        end
+
         # Submit button
         #
         # @param content [String] The content
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -1188,4 +1188,56 @@ describe Hanami::Helpers::FormHelper do
       end
     end
   end
+
+  describe "#datalist" do
+    let(:values) { Hash['it' => 'Italy', 'us' => 'United States'] }
+
+    it "renders" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="book[store]" id="book-store" list="books">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
+    end
+
+    it "just allows to override 'id' attribute of the text input" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books', id: 'store'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="book[store]" id="store" list="books">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
+    end
+
+    it "allows to override 'name' attribute" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books', name: 'store'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="store" id="book-store" list="books">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
+    end
+
+    it "allows to specify HTML attributes" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books', class: 'form-control'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="book[store]" id="book-store" list="books" class="form-control">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
+    end
+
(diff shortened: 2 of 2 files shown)


AlfonsoUceda:
@hoksilato thanks for your PR!

I think datalist method does two things, it creates input and datalist tags. You can create inputs with the current html builder implementation. So I'd modify the method for creating only datalist instead the two types.

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -766,6 +766,45 @@ def select(name, values, attributes = {})
...
+        #   #    <option value="Italy"></option>
+        #   #    <option value="United States"></option>
+        #   #  </datalist>
+        def datalist(name, values, list, attributes = {})
+          options  = attributes.delete(:options) || {}
+          datalist = attributes.delete(:datalist) || {}
+          attributes = { type: 'text', name: _input_name(name), id: _input_id(name), list: list }.merge(attributes)
+
+          # remove 'id' attribute
+          datalist = { id: list }.merge(datalist)
```
davydovanton: I think that `#[]=` method will be better here, i.e:
``` ruby
 datalist[id] = list
```

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -766,6 +766,45 @@ def select(name, values, attributes = {})
...
+
+          input(attributes)
+          super(datalist) do
+            values.each do |value, content|
+              option(value, {value: content}.merge(options))
```
davydovanton: and here too. What do you think?

lib/hanami/helpers/form_helper/form_builder.rb
hoksilato: Yes, it looks better. Thanks.

hoksilato:
@AlfonsoUceda if I separate the `input` and `datalist`, I have to make sure the `list` attribute of `input` will be same with `id` attribute of `datalist`. Beside that `datalist` alone is not make sense. It's my IMHO.

If you see it's fine, I'd modify as your suggestion :grinning:

lib/hanami/helpers/form_helper/form_builder.rb
hoksilato: Sorry, I can't understand what you mean, could you please explan. Thanks.

joneslee85: @hoksilato I think @davydovanton meant that you should replicate the pattern of mutating hash using `Hash#[]` as aforementioned

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -766,6 +766,45 @@ def select(name, values, attributes = {})
...
 
+        # Datalist input
+        #
+        # @param name [Symbol] the input name
+        # @param values [Hash] a Hash to generate <tt><option></tt> tags of datalist.
+        #   Keys correspond to <tt>value</tt> and values correspond to the content.
+        # @param list [String] the name of list for the text input, it's also the id of datalist
+        # @param attributes [Hash] HTML attributes to pass to the input tag
+        #
+        # @since 0.3.1
```
joneslee85: s/0.3.0/x.x.x

test/form_helper_test.rb
```diff
@@ -1188,4 +1188,56 @@
...
+
+    it "renders" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="book[store]" id="book-store" list="books">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
+    end
+
+    it "just allows to override 'id' attribute of the text input" do
```
joneslee85: s/just//

lib/hanami/helpers/form_helper/form_builder.rb
davydovanton: yes, @joneslee85 right. I want to say that this code:
``` ruby
options[:value] = content
option(value, options)
```
will be faster than current solution.

test/form_helper_test.rb
```diff
@@ -1188,4 +1188,56 @@
...
+
+  describe "#datalist" do
+    let(:values) { Hash['it' => 'Italy', 'us' => 'United States'] }
+
+    it "renders" do
+      actual = view.form_for(:book, action) do
+        datalist :store, values, 'books'
+      end.to_s
+
+      actual.must_include %(<input type="text" name="book[store]" id="book-store" list="books">\n<datalist id="books">\n<option value="Italy">it</option>\n<option value="United States">us</option>\n</datalist>)
```
joneslee85: IMHO the key should be the text node and the value should be the `value` attribute.

```
<option value="United States">us</option>
# => TO
<option value="us">United States</option>
```

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -766,6 +766,45 @@ def select(name, values, attributes = {})
...
+        #   <%=
+        #     # ...
+        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     datalist :stores, values, 'books'
+        #   %>
+        #
+        #   # Output:
+        #   #  <input type="text" name="book[store]" id="book-store" list="books">
+        #   #  <datalist id="books">
+        #   #    <option value="Italy"></option>
```
joneslee85: seems to me the value missing the text node

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -766,6 +766,45 @@ def select(name, values, attributes = {})
...
+        #
+        # @example Basic usage
+        #   <%=
+        #     # ...
+        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     datalist :stores, values, 'books'
+        #   %>
+        #
+        #   # Output:
+        #   #  <input type="text" name="book[store]" id="book-store" list="books">
```
joneslee85: I think we should discuss about the generation if `<input>` tag. To me it makes the helper complex, options like `options`, `datalist` to manipulate the output are clunky compare to explicit and verbose declaration of nested `input` then `datalist` tags

lib/hanami/helpers/form_helper/form_builder.rb
hoksilato: Thanks

test/form_helper_test.rb
hoksilato: If we do that the options won't display right. You can see at [here](https://jsbin.com/yutufobizu/edit?html,output)

jodosha:
@hoksilato Thanks for this implementation. I just merged it :+1:

hoksilato:
@jodosha thanks ^^

Issue #59, opened 2016-03-02
Country select helper


cllns:
[`country_select`](https://github.com/stefanpenner/country_select) is a useful Rails gem to generate a `select` of countries, with their country codes as values. I'm working on a `hanami` app right now that could benefit from it.

Can this be ported to Hanami? Note: Rails removed it from the `rails` project for political reasons: the names/existence of countries can be controversial. Hanami may wish to avoid that too, though going with ISO-3166 seems like a relatively fair option.

I'm happy to either build it into `hanami/helpers` or make my own gem for it.

jodosha:
@cllns This is a good idea. :+1: 

What we aim for, is to NOT create a walled garden for Hanami Community, but something that all the Ruby ecosystem can benefit from.

For instance, all the `select` helpers across Hanami, Rails and Padrino accept an enumerable (Hash, or nested arrays). It would be nice if the first goal of your gem is to provide that basic data structure. Eg:

```ruby
module Countries
  def self.mapping
    { "it" => "Italy" }
  end
end
```

This is just Ruby, so all the Ruby projects can use this data structure to build their country select.

---

On the other hand, you can offer out-of-the-box support for Hanami. Eg.

```ruby
begin
  require 'hanami/helpers'
rescue LoadError
end

if defined?(Hanami::Helpers)
  # ...
end
```

WDYT?

PR #61, opened 2016-03-07
Invert values for form builder's select

jodosha:
**This is a breaking change proposal. If approved, we have to bump the minor version to `v0.4.0`.**

---

## Actual

The `values` has must to be composed with: keys as `value` attribute and values as content of the `<option>` tag.

### Example

```erb
<%=
  values = Hash['it' => 'Italy', 'us' => 'United States']
  select :stores, values
%>
```

```html
<select name="book[store]" id="book-store">
  <option value="it">Italy</option>
  <option value="us">United States</option>
</select>
```

## Proposed

The `values` has must to be composed with: values as `value` attribute and keys as content of the `<option>` tag.

### Example

```erb
<%=
  values = Hash['Italy' => 'it', 'United States' => 'us']
  select :stores, values
%>
```

```html
<select name="book[store]" id="book-store">
  <option value="it">Italy</option>
  <option value="us">United States</option>
</select>
```

### Reasons

There are a few reasons:

  * Better "sortable" experience. Sorting a Hash by key, is easier than by value. Ref https://github.com/hanami/helpers/issues/59
  * We had to implement `datalist` helper by using this new "inverted" syntax. Ref https://github.com/hanami/helpers/pull/60 and https://github.com/hanami/helpers/commit/4a885a6699892ac78b74ad08f7cc04b9b89bc4dd
  * It's consistent across Ruby ecosystem: Rails and Padrino this "inverted" syntax. This avoids the mental overhead of switching between frameworks

---

/cc @hanami/core-team @hanami/contributors

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -711,7 +711,7 @@ module Hanami
         #
         # @param name [Symbol] the input name
         # @param values [Hash] a Hash to generate <tt><option></tt> tags.
-        #   Keys correspond to <tt>value</tt> and values correspond to the content.
+        #   Values correspond to <tt>value</tt> and keys correspond to the content.
         # @param attributes [Hash] HTML attributes to pass to the input tag
         #
         # If request params have a value that corresponds to one of the given values,
@@ -723,7 +723,7 @@ module Hanami
         # @example Basic usage
         #   <%=
         #     # ...
-        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     values = Hash['Italy' => 'it', 'United States' => 'us']
         #     select :stores, values
         #   %>
         #
@@ -756,7 +756,7 @@ module Hanami
           attributes = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
 
           super(attributes) do
-            values.each do |value, content|
+            values.each do |content, value|
               if _value(name) == value
                 option(content, {value: value, selected: SELECTED}.merge(options))
               else
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -3,6 +3,12 @@ require 'hanami/controller'
 require 'hanami/helpers/html_helper'
 require 'hanami/helpers/escape_helper'
 
+Store = Struct.new(:code, :label) do
+  def to_ary
+    [label, code]
+  end
+end
+
 class HtmlView
   include Hanami::Helpers::HtmlHelper
 
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -1133,7 +1133,7 @@ describe Hanami::Helpers::FormHelper do
   end
 
   describe "#select" do
-    let(:values) { Hash['it' => 'Italy', 'us' => 'United States'] }
+    let(:values) { Hash['Italy' => 'it', 'United States' => 'us'] }
 
     it "renders" do
       actual = view.form_for(:book, action) do
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -1175,6 +1175,56 @@ describe Hanami::Helpers::FormHelper do
       actual.must_include %(<select name="book[store]" id="book-store">\n<option value="it" class="form-option">Italy</option>\n<option value="us" class="form-option">United States</option>\n</select>)
     end
 
+    describe "with values an structured Array of values" do
+      let(:values) { [['Italy', 'it'], ['United States', 'us']] }
+
+      it "renders" do
+        actual = view.form_for(:book, action) do
+          select :store, values
+        end.to_s
+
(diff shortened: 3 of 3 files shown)


pascalbetz:
How would ```optgroups``` be modeled?

davydovanton:
I don't like this because for me `{ en: 'English language' }` is better than `{ 'English language' => 'en' }`. But anyway this changes will avoid the mental overhead of switching between frameworks. That's why I think that it good proposal :+1:

jodosha:
@pascalbetz We should probably create a **new helper for that**:

```erb
<%=
  optgroup Hash[
    "Group 1" => { "Option 1.1" => "1.1" },
    "Group 2" => { "Option 2.1" => "2.1", "Option 2.2" => "2.2" }
  ]
%>
```

```html
  <optgroup label="Group 1">
    <option value="1.1">Option 1.1</option>
  </optgroup> 
  <optgroup label="Group 2">
    <option value="2.1">Option 2.1</option>
    <option value="2.2">Option 2.2</option>
  </optgroup>
```

Then we should make `select` and `datalist` to accept this kind of input:

```erb
<%=
  values = optgroup Hash[
    "Group 1" => { "Option 1.1" => "1.1" },
    "Group 2" => { "Option 2.1" => "2.1", "Option 2.2" => "2.2" }
  ]

  select :foo, values
%>
```


```html
<select>
  <optgroup label="Group 1">
    <option value="1.1">Option 1.1</option>
  </optgroup> 
  <optgroup label="Group 2">
    <option value="2.1">Option 2.1</option>
    <option value="2.2">Option 2.2</option>
  </optgroup>
</select>
```

WDYT?

jodosha:
@davydovanton 
> I don't like this because for me `{ en: 'English language' }` is better than `{ 'English language' => 'en' }`.

Yes, that is the reason why it's implemented as it is.

davydovanton:
@jodosha also I had a problems with getting options value by key (when key was long string or smth like this) :disappointed:

jodosha:
@davydovanton 
> also I had a problems with getting options value by key (when key was long string or smth like this)

Most of the time you work with variables, for options, so you don't see actual values to clutter source code. Am I missing something?

davydovanton:
@jodosha nope, all as you said

pascalbetz:
@jodosha

I don't like the change. Does not feel the right way...but i can live with it:-)
Since with an additional helper for ```optgroups``` we would need to add logic to ```select``` and ```datalist``` anyway (detect if options is a string or a hash) i'd suggest to make ```select``` and ```datalist``` smart enough to also accept a nested structure and build options accordingly. 


I think the helpers should also (or only) work with ```Array```:

```ruby
options = [['Italian', 'it'], [['German', 'de']], ...]
```
This is easier/more straightforward to build using ```map```. For a hash we'd need to use ```inject``` or ```each_with_object```. I often need this for options that are built depending on some state in the DB.

```ruby
Language = Struct.new(:code, :label)


LANGUAGES = [Language.new(:it, 'Italian'), Language.new(:de, 'German')]


options_from_map = LANGUAGES.map do |language|
  [language.label, language.code]
end

options_from_inject = LANGUAGES.inject({}) do |memo, language|
  memo[language.label] = language.code
  memo
end

options_from_each_with_object = LANGUAGES.each_with_object({}) do |language, memo|
  memo[language.label] = language.code
end
```

So:

* Switch the order if we must
* Make it accept Array
* Make it accept Hash
* Make it accept nested structure


Cheers.

jodosha:
@pascalbetz Thanks for your input. With https://github.com/hanami/helpers/commit/0f74402536aee78004d9f645081bb9db7d34b403 we introduced new capabilities:

```erb
<%=
  values = [['Italy', 'it'], ['United States', 'us']]
  select :stores, values
%>
```

and

```ruby
Store = Struct.new(:code, :label) do
  def to_ary
    [label, code]
  end
end
```

```erb
<%=
  values = [Store.new('it', 'Italy'), Store.new('us', 'United States')]
  select :stores, values
%>
```

jodosha:
@pascalbetz 
> Make it accept nested structure

Is this a reference to `optgroup`? We should probably introduce it in a new PR.

pascalbetz:
@jodosha yes, the "nested structure" would be about ```optgroup```.

rpanachi:
I can't understand the motivation for this "enhancement". I updated Hanami from `v0.7` to `v0.8` and now I need to change all my forms to this inverted key/value options. What the value that this change made to Hanami as a web framework?

jodosha:
@rpanachi I'm sorry you had to manually update your forms. It sucks, I know.

But please understand that getting all the API right at the first shot is impossible for a large project like Hanami. We took so long to reach 1.0 (we're almost there), to guarantee that the API won't change after this important milestone.

Thanks for your patience.

---

The motivations are explained in the PR description. In a nutshell: it's for compatibility with the Ruby ecosystem and because our new `datalist` helper needed that syntax. So for consistency sake we flipped the syntax here too.

PR #62, opened 2016-03-10
Explicitly require some hanami-utils dependencies

killthekitten:
Thank you for such an awesome tool. I'm using `hanami-helpers` within a `middleman` project, and occasionally ran into dependency issues with some of the helpers.


jodosha:
@killthekitten Thanks for this improvement. :+1:

Issue #66, opened 2016-06-07
the selected attribute on the option tag is not set


sebastjan-hribar:
Hi,

I'm using select helper and the selected attribute on the option tag is not automatically set. For example, when displaying an entity for edit and update.

I've setup a small bookshelf with book and author entities. It's minimalistic just to get the minimum required environment ready.

Hanami data:
hanami,       '0.7.2'
hanami-model, '~> 0.5' 
hanami-helpers (~> 0.3)

Sample app: https://github.com/sebastjan-hribar/hanami-bookshelf

This is the html from inspecting page:

```
<h2>Edit book</h2>

<form action="/books/1" method="POST" accept-charset="utf-8" id="book-form">
<input type="hidden" name="_method" value="PATCH">
<div class="input">
<label for="book-title">Title</label>
<input type="text" name="book[title]" id="book-title" value="Test Book">
</div>
<div class="input">
<label for="book-author">Author</label>
<select name="book[author]" id="book-author">
<option value="1">Best Author</option>
<option value="2">Second Best Author</option>
</select>
</div>
<div class="controls">
<button type="submit">Update Book</button>
</div>
</form>
```

I've similar problems in my other app, I've setup this one just for this select problem.

Please let me know if I should add further information.

regards,
Sebastjan

jodosha:
@sebastjan-hribar Thanks for reporting this problem.

I've cloned your demo app (thanks for that 🙏 ). It can't work, because the framework only selects an `<option>` tag in case of a HTTP param has a value that applies. This is useful in case of failing form submission: you want to render the form again, but won't loose the input that the user typed.

We should expand `select` with a new option to cover your case.

```erb
select :author, authors, selected: book.id
```

Then this check should be expanded accordingly: https://github.com/hanami/helpers/blob/4e91ee6fe1aa1096fa2484f90d159ef283433b21/lib/hanami/helpers/form_helper/form_builder.rb#L779

WDYT? Is it something that you want to implement? Thanks!

sebastjan-hribar:
@jodosha I'm sorry for my late reply.

Thank you for the feedback.

`select :author, authors, selected: book.id` is something I had in mind for this scenario.

I'd be interested in trying to implement this, however I'm not confident I can do as good a job and as fast as an experienced person. But I have to start somewhere and sometime, right? :smile: 
Let me know if you would like to entrust this to someone else, or I should have a go at it.

On a second note:
[...]

PR #69, opened 2016-06-27
Set selected option based on entity's attribute

sebastjan-hribar:
This is the implementation for https://github.com/hanami/helpers/issues/66.

It enables to set the selected option if one of the option values matches the passed entity's attribute:

`select :author, authors, selected: book.id`


jodosha:
@sebastjan-hribar This looks good. Thanks!

Issue #71, opened 2016-07-28
form_for helper giving an undefined method 'params' in tests


TiteiKo:
Hi,

So.. I'm trying to make a really simple form:

```ruby
# in the view
def creation_form
  Form.new(:thing, '#', thing: thing) # thing being exposed by the action
end

def creation_form_submit_text
  'Create'
end
```

```haml
-# and the template (in haml)
= form_for creation_form, class: '' do
  - text_field :title
  - submit creation_form_submit_text
```

With a really small test on the view/template:

```ruby
require 'spec_helper'
require_relative '../../../../apps/web/views/things/new'

describe Web::Views::Things::New do
  let(:exposures) { Hash[thing: Thing.new] }
  let(:template)  { Hanami::View::Template.new('apps/web/templates/things/new.html.haml') }
  let(:view)      { Web::Views::Things::New.new(template, exposures) }
  let(:rendered)  { view.render }

  it('renders') { rendered }
end
```

And everything should be shining in glory to the Thing! And it is! At least in my browser, yay!

But it's not in my test...

```
NoMethodError: undefined method `params' for #<Hanami::View::Rendering::Scope:0x007ff49c85a9b0>
```

Full error stack: http://0bin.net/paste/qIW55zzj6-TmTtCR#gm-ZUExIHvX8PhTJi0bZMPTHPUI0qN3hu/KMLSsyjv/
Full Gemfile.lock: http://0bin.net/paste/2RiASve4iefyZ+Co#UrbYcXYSbimlQwyRqba1vviD93uD+ql7KWx30cJ6mrE

bruz:
I ran into this too at one point. Adding `params` to the exposures seems to fix it, so you could try:

```ruby
require 'spec_helper'
require_relative '../../../../apps/web/views/things/new'

class NewThingsParams < Hanami::Action::Params
  params do
  end
end

describe Web::Views::Things::New do
  let(:params)    { NewThingsParams.new({}) }
  let(:exposures) { Hash[thing: Thing.new, params: params] }
  let(:template)  { Hanami::View::Template.new('apps/web/templates/things/new.html.haml') }
  let(:view)      { Web::Views::Things::New.new(template, exposures) }
  let(:rendered)  { view.render }

  it('renders') { rendered }
end
```

I noticed that as of Hanami 0.8.0 I had to add the `params` block in `NewThingsParams` to get it to work.

If someone knows a less verbose way of doing setup for view tests, I would love having a better way of doing this.

[...]

PR #72, opened 2016-08-12
Form values to support `Hash#dig`

jodosha:
This is useful to support unit tests setup in Hanami projects, when
developers wants to pass pure Hash to simulate params, instead of an
instance of `Hanami::Action::Params`.

Ref https://github.com/hanami/helpers/issues/71#issuecomment-239104711

### DIFF
--- a/lib/hanami/helpers/form_helper/values.rb
+++ b/lib/hanami/helpers/form_helper/values.rb
@@ -67,7 +67,7 @@ module Hanami
         # @api private
         def _get_from_params(key)
           if @params.respond_to?(:dig)
-            @params.dig(*key.to_s.split(GET_SEPARATOR).map(&:to_sym))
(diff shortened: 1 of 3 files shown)


lib/hanami/helpers/form_helper/values.rb
```diff
@@ -27,11 +27,52 @@ def initialize(values, params)
...
+        # To make this possible, we should change the key notation from a dot
+        # separated string to an array of symbol.
+        #
+        # FIXME: Use Hash#dig when we'll support only Ruby 2.3+
+        #
+        # @since x.x.x
+        # @api private
+        def _get_from_params(key)
+          if @params.respond_to?(:dig)
+            @params.dig(*key.to_s.split(GET_SEPARATOR).map(&:to_sym))
```
davydovanton: little micro optimization: I think `#map!` will be better here, what do you think?

joneslee85:
:+1:

PR #73, opened 2016-08-29
Checkbox behaves unexpectedly when passing a boolean in values

ksenia-zalesnaya:
With this test I get an error 

```
NoMethodError: undefined method `include?' for true:TrueClass
    helpers/lib/hanami/helpers/form_helper/form_builder.rb:1036:in `_attributes_for_check_box'
```

I guess this include? is for multiple checkboxes, it also works with string values like '1' or 'true' from params, but I pass model with boolean attributes to form. May be there should be a check that value is an array?

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -678,7 +678,7 @@ module Hanami
         #   #  <input type="radio" name="book[category]" value="Non-Fiction" checked="checked">
         def radio_button(name, value, attributes = {})
           attributes = { type: :radio, name: _input_name(name), value: value }.merge(attributes)
-          attributes[:checked] = CHECKED if _value(name) == value
+          attributes[:checked] = CHECKED if _value(name).to_s == value.to_s
           input(attributes)
         end
 
@@ -1050,7 +1050,7 @@ module Hanami
         # rubocop:disable Metrics/PerceivedComplexity
         def _select_option_selected?(value, selected, input_value, multiple)
           value == selected || (multiple && (selected.is_a?(Array) && selected.include?(value))) ||
-            value == input_value || (multiple && (input_value.is_a?(Array) && input_value.include?(value)))
+            value.to_s == input_value.to_s || (multiple && (input_value.is_a?(Array) && input_value.include?(value)))
         end
         # rubocop:enable Metrics/PerceivedComplexity
         # rubocop:enable Metrics/CyclomaticComplexity
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1030,10 +1030,7 @@ module Hanami
             value: attributes.delete(:checked_value) || DEFAULT_CHECKED_VALUE
           }.merge(attributes)
 
-          value = _value(name)
-          attributes[:checked] = CHECKED if !value.nil? &&
-                                            (value.to_s == attributes[:value].to_s || value.is_a?(TrueClass) ||
-                                             value.is_a?(Array) && value.include?(attributes[:value]))
+          attributes[:checked] = CHECKED if _check_box_checked?(attributes[:value], _value(name))
 
           attributes
         end
@@ -1054,6 +1051,12 @@ module Hanami
         end
         # rubocop:enable Metrics/PerceivedComplexity
         # rubocop:enable Metrics/CyclomaticComplexity
+
+        def _check_box_checked?(value, input_value)
+          !input_value.nil? &&
+            (input_value.to_s == value.to_s || input_value.is_a?(TrueClass) ||
+            input_value.is_a?(Array) && input_value.include?(value))
+        end
       end
     end
   end
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -678,7 +678,7 @@ module Hanami
         #   #  <input type="radio" name="book[category]" value="Non-Fiction" checked="checked">
         def radio_button(name, value, attributes = {})
           attributes = { type: :radio, name: _input_name(name), value: value }.merge(attributes)
-          attributes[:checked] = CHECKED if _value(name) == value
+          attributes[:checked] = CHECKED if _value(name).to_s == value.to_s
           input(attributes)
         end
 
@@ -1032,7 +1032,7 @@ module Hanami
 
           value = _value(name)
           attributes[:checked] = CHECKED if !value.nil? &&
-                                            (value == attributes[:value] ||
+                                            (value.to_s == attributes[:value].to_s ||
                                              value.is_a?(Array) && value.include?(attributes[:value]))
 
           attributes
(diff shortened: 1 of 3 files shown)


cllns:
Thanks @ksenia-zalesnaya!

Should we be allowing `true` for `checked_value` and `unchecked_value`? I'd almost rather throw an error and tell the user to send a string instead, either `true.to_s` or `"true"`. Helps the user remember (or learn) that a string will be submitted from a form, not a boolean. 

If we do want to allow booleans though, then we could be more flexible by calling `.to_s` on the values to see if they're equal first:   
```
(value.to_s == attributes[:value].to_s ||
 value.include?(attributes[:value]))
```
The `checked_value` could be `'true'` and if the underlying value is `true`, it'd still work. (And vice versa, though that seems less likely). Would also allow symbols.

ksenia-zalesnaya:
@cllns I think we should allow booleans. If model comes from database it is likely to have boolean and integer values and it would be annoying making them all strings manually. Also I think we should keep my fix with array, because include? is for multiple checkboxes and has no sense with other types. (What if value and attributes[:value] were matching strings? Matching partially is not an expected behaviour.)

ksenia-zalesnaya:
Encountered similar problem with radio_button, had to write explicitly `radio_button :object_id, object.id.to_s`. I suppose to_s should be done in all checks like `_value(name) == value`.
And maybe not concerning this issue, but also about form values - shouldn't hashes be allowed as values as well as objects?

cllns:
Yes, your point about wanting to only use `.include?` on Arrays and not Strings (or non-Array Enumerables) is a good one.

I just tried running test suite with the test you added, but reverting the code change, and it passed for me.


Here are some test I wrote for the `.to_s` thing (the two bottom ones fail without my code change above):

```
# Line 291
    describe 'automatic values' do
      describe 'value boolean, helper boolean' do
        let(:values) { Hash[book: OpenStruct.new(free_shipping: true)] }

        it "renders with 'checked' attribute" do
          actual = view.form_for(:book, action, values: values) do
            check_box :free_shipping, checked_value: true
          end.to_s

          actual.must_include %(<input type="checkbox" name="book[free_shipping]" id="book-free-shipping" value="true" checked="checked">)
        end
      end

      describe 'value boolean, helper string' do
        let(:values) { Hash[book: OpenStruct.new(free_shipping: true)] }

        it "renders with 'checked' attribute" do
          actual = view.form_for(:book, action, values: values) do
            check_box :free_shipping, checked_value: 'true'
          end.to_s

          actual.must_include %(<input type="checkbox" name="book[free_shipping]" id="book-free-shipping" value="true" checked="checked">)
        end
      end

      describe 'value string, helper boolean' do
        let(:values) { Hash[book: OpenStruct.new(free_shipping: 'true')] }

        it "renders with 'checked' attribute" do
          actual = view.form_for(:book, action, values: values) do
            check_box :free_shipping, checked_value: true
          end.to_s

          actual.must_include %(<input type="checkbox" name="book[free_shipping]" id="book-free-shipping" value="true" checked="checked">)
        end
      end
    end
  end
```
We'd also have to make sure this is consistent for other helpers, though not sure if this change would apply to any of them.

I think hashes as values would be a good separate issue :)

ksenia-zalesnaya:
I am not sure how I got it falling last time. Now it falls when values are not equal booleans.
Also I found some `.to_s` cases in setting values from params (that's what I wrote about radio button earlier).
For checkbox setting true as checked value results either in not setting value from model (where it is boolean), or from params (where it is string). Added params test for it.

TiteiKo:
We could simplify the use by doing something like this:

```ruby
attributes[:checked] = CHECKED if !value.nil? &&
                                   (value == attributes[:value] || value.is_a?(TrueClass) ||
                                   (value.is_a?(Array) && value.include?(attributes[:value])))
```

This solves the problem while staying user friendly as it will work out of the box for booleans

TiteiKo:
See https://github.com/hanami/helpers/compare/master...TiteiKo:fix-checkboxes-for-booleans?expand=1

jodosha:
@ksenia-zalesnaya Hi and thanks for this fix. The build is failing both locally and CI. Can you please check it? Thanks!

AlfonsoUceda:
ping @ksenia-zalesnaya

ksenia-zalesnaya:
@AlfonsoUceda sorry, I'm on vacation, I'll take a look the next week

AlfonsoUceda:
thanks @ksenia-zalesnaya

ksenia-zalesnaya:
As far as I remember I wrote falling tests intentionally to show cases in which I had problems. But I didn't know what is the best way to fix them. I don't think we should explicitly check for booleans like  @TiteiKo suggested - it's OK with checkboxes, but with radiobuttons we may also have integer and decimal values. So the easiest way seems to cast to string. Maybe you have better ideas?

jodosha:
@TiteiKo Do you want to discuss with @ksenia-zalesnaya to find a solution together? Thank you!

TiteiKo:
@ksenia-zalesnaya I think your PR covers more than the initial problem, which was only on the checkboxes. We should focus on this problem here, and radio buttons problem in another PR, as they are different topics.

The problem at hand is that a checkbox populated by a boolean fails because `TrueClass` doesn't respond to `include?`.

---

```ruby
(value.to_s == attributes[:value].to_s ||
  value.is_a?(Array) && value.include?(attributes[:value]))
```

I agree on comparing after casting the values to strings, and ofc on the check to see if the value is an Array.

---

It fixes the problem.

But in the context of a boolean, we have to choose between two ways of writing the helper in the form:
```ruby
check_box :free_shipping, checked_value: true
# or
check_box :free_shipping
```

> The input element represents a two-state control that represents the element's checkedness state. If the element's checkedness state is true, the control represents a positive selection, and if it is false, a negative selection. If the element's indeterminate IDL attribute is set to true, then the control's selection should be obscured as if the control was in a third, indeterminate, state.
  https://www.w3.org/TR/html5/forms.html#checkbox-state-(type=checkbox)

Having the check_value being `true` (or a common representation of `true`as `1`) is what most people are used to. I don't think they should have to always specify that `checked_value: true`

Hence in my branch the added 
```ruby
|| value.is_a?(TrueClass)
```

Downside of this addition: If someone wanted to use `false` as the checked value and `true` as the unchecked value, the checkbox would always be checked. _(but that's a really edgy abd uncommon use case...)_

---

The solution I think would be best is a mix between @ksenia-zalesnaya and mine:

```ruby
if !value.nil? &&
  (value.to_s == attributes[:value].to_s || value.is_a?(TrueClass) ||
  value.is_a?(Array) && value.include?(attributes[:value]))
```

As the `_attributes_for_check_box` method is only used in the checkbox helper, it won't change anything for radio buttons. But as I said in the beginning of this way too long answer, one PR for one problem.

ksenia-zalesnaya:
@TiteiKo I see your point. Created separate [PR](https://github.com/hanami/helpers/pull/73) for radio and selects.

TiteiKo:
@ksenia-zalesnaya Could you clean up the tests and rebase your branch please?

ksenia-zalesnaya:
I made plain old rebase, and git shows only my commits in diff between branches and no conflicts, but here it looks like I did something wrong, did I?

ksenia-zalesnaya:
@TiteiKo Everything seems to be OK now?

TiteiKo:
@ksenia-zalesnaya It looks great thanks! Last little thing and I'll merge: could you squash your commits please?

PR #74, opened 2016-08-30
add form support for arrays of nested fields

TiteiKo:
Hi all,

~~This still needs a bit of work, especially on the naming domain. It also needs https://github.com/hanami/controller/pull/177 to be merged (and the Gemfile updated)~~
#### What it does

At the moment, we're able to build forms with nested parameters, as long as they're vertically flat:

``` ruby
foo: { bar: { baz: 'foobar' } }
```

This PR allows us to build forms for structures like the following:

``` ruby
foo: { bar: [{ baz: 'foobar' }, { baz: 'barfoo' }]}
```
#### Things I'm not happy about in this PR (but need some help figuring out):
- ~~the name `structured_fields_for`... I'm so damn bad at naming...~~
- ~~having a `_cleaned_input_name` method.. We need to have uniq fields ids (ie: `array-0-field`), and we need to have the value key with the index (ie: `array.0.field`), but rack needs the field name without index (ie: `array[][field]`). But it should be the other way around, having a field full name used for id and value, and the field name being extrapolated from this full name.~~
#### ~~Note on the failing test~~

~~The test not passing is the one already not passing with https://github.com/hanami/controller master branch (see https://github.com/hanami/controller/issues/176 )~~


jodosha:
@TiteiKo Hi and thanks for this PR. Why `fields_for` isn't enough for your case? Is just a matter to populate an array of fields with an array of data coming from values/params?

AlfonsoUceda:
ping @TiteiKo

TiteiKo:
Hi @jodosha. It's not only populating from an array, it's also computing the right input name, id and value. The aim is to support sub structures (for example has_many associations)

Basically, the following should happen:

```ruby
# with the following data
{foo: { bar: [{ baz: 'foobar' }, { baz: 'barfoo' }]}}

# and the following form
form_for :foo, url: foo_path do 
  name_it_what_you_want :bar do |index|
    text_field :baz
  end
end

# there should be two inputs
input id: 'foo-bar-0-baz', name: 'foo[bar][][baz]', value: 'foobar'
input id: 'foo-bar-1-baz', name: 'foo[bar][][baz]', value: 'barfoo'
```

AFAIK, `form_for` can't loop through an array, and it shouldn't, it's two different jobs, and it's the dev's job to know the relation between the structure and sub-structure.

jodosha:
@TiteiKo Is it fine with you if we hold on with this? I'm still not sure about the feature, the naming and the implementation.

@AlfonsoUceda WDYT?

AlfonsoUceda:
@TiteiKo can't you insert the following data structure with the current approach?
```ruby
{foo: { bar: [{ baz: 'foobar' }, { baz: 'barfoo' }]}}
```

I mean, imagine you have an array of objects and you want to use in `fields_for`.

TiteiKo:
@AlfonsoUceda No you can't, that's why I did this branch
@jodosha No problem, I'm not satisfied with the name or the implementation either, but the feature itself is a real need for advanced projects.

I'll rebase my branch and clean it up a bit

AlfonsoUceda:
@TiteiKo then I think we should work for amending `fields_for` to support that case. what do you think folks?

mereghost:
*uses his dreaded PR necromancy* ping @TiteiKo 

My only fear of including this use case on fields_for is getting it a bit to complex, but that's pure baseless fear. =p

TiteiKo:
@mereghost I was waiting for others to give their opinion ;)

Including this use case in fields_for would be a bad idea:

* increased complexity
* mix up between what should be submitted as an array (of schemas) vs what should be submitted as a schema
* either no possibility to have the current index when in the loop, or having to deal with it when there's no index because there's no loop because it's not an array

I'm open to suggestions to improve this PR, as I said, the naming is terrible and the code is not that great either. But I still think it should be a new method for this use-case.

test/fixtures/templates/bills/edit.html.erb
```diff
@@ -0,0 +1,18 @@
...
+      legend 'Addresses'
+
+      fields_for_collection :addresses do |i|
+        div class: 'form-group' do
+          label :street
+          input_text :street, class: 'form-control', placeholder: 'Street', 'data-funky': "id-#{i}"
+        end
+      end
+
+      label :ensure_names
```
jodosha: You remembered about this, great! 😉 👍

TiteiKo:
So, here are the changes @jodosha and I agreed last week:

* we keep the method, as it's different from `#fields_for` and would greatly increase the complexity of `#fields_for` if we were to have only one method for both use-cases.
* we've renamed the method to `#fields_for_collection`. If you don't like it and have a better proposal, feel free to say so before we hit the first RC :) 

Sidenotes: 

* I've rebased the branch on master and the tests fails to install the rainbow gem :(
* No test or rubocop errors on my computer

jodosha:
@TiteiKo For the rainbow gem, here's how to fix it: https://github.com/sickill/rainbow/issues/44#issuecomment-274412027

jodosha:
@TiteiKo You can merge this 👍

AlfonsoUceda:
@TiteiKo could you rebase this PR? thanks!

AlfonsoUceda:
Merged!

PR #76, opened 2016-09-09
fix failing test with controller master branch

TiteiKo:
See https://github.com/hanami/controller/issues/176


TiteiKo:
See https://github.com/hanami/helpers/pull/83 since https://github.com/hanami/helpers/commit/977cf65eec9837ab6ececf36195ec4a6f8b7fd33

PR #79, opened 2016-10-06
Fix test data for html form generator test

davydovanton:


TiteiKo:
@hanami/core-team I think this PR can be closed as the fix was made and merged here https://github.com/hanami/helpers/pull/83

jodosha:
@davydovanton Hi, and thank you! But @TiteiKo is right: her PR fixes the problem the right way.

If you look at the [test description](https://github.com/hanami/helpers/blob/2238cd8dbdc7c95663b14447719712e659a0b3ba/test/integration/form_helper_test.rb#L79) it says:

```ruby
    describe 'after a failed submission with blank values' do
```

So that value **has** to be blank string. Not even `nil`, as I wrongly did in the past.

davydovanton:
@jodosha okay, thanks for explanation 👍

PR #81, opened 2016-10-10
add multiple option possibility to select

TiteiKo:
We can't ATM have a fully working select that can select multiple values.

This PR adds this use case to the `select` helper

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -780,9 +780,22 @@ module Hanami
         #   #    <option value="it" selected="selected">Italy</option>
         #   #    <option value="us">United States</option>
         #   #  </select>
+        #
+        # @example Multiple select
+        #   <%=
+        #     # ...
+        #     values = Hash['it' => 'Italy', 'us' => 'United States']
+        #     select :stores, values, multiple: true
+        #   %>
+        #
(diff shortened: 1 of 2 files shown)


lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -780,17 +780,30 @@ def password_field(name, attributes = {})
...
+        #
+        #   # Output:
+        #   # <select name="book[store][]" id="book-store" multiple="multiple">
+        #   #   <option value="it">Italy</option>
+        #   #    <option value="us">United States</option>
+        #   #  </select>
         def select(name, values, attributes = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
           options    = attributes.delete(:options) { {} }
-          attributes = { name: _input_name(name), id: _input_id(name) }.merge(attributes)
+          attributes = { name: _select_input_name(name, attributes[:multiple]), id: _input_id(name) }.merge(attributes)
```
beauby: How about the following? (to avoid unnecessary computations and object instantiations)
```ruby
attributes[:name] ||= _select_input_name(name, attributes[:multiple])
attributes[:id] ||= _input_id(name)
```

jodosha:
@TiteiKo Hi and thanks! This looks good: merging.

PR #83, opened 2016-11-04
Fix blank submission test

TiteiKo:
This test is about re-rendering the form when a field was emptied.

* the parameter sent by rack would be an empty string and not nil
* the form should not re-fill the field (especially if it's not the field causing the error)


jodosha:
@TiteiKo Thanks for fixing my mistake! 🙌 👍

Issue #77, opened 2016-09-09
Number formatting precision


TiteiKo:
Hi all,

It looks like `NumberFormattingHelper` does not fully take into account the given precision:

``` ruby
format_number 3.14, precision: 0
# => 3.0 (instead of 3)
format_number 3.14, precision: 4
# => 3.14 (instead of 3.1400)
```

Is this by design?

katafrakt:
Current implementation is based on Ruby `round` method, which behaves like that. See: https://ruby-doc.org/core-2.3.0/Float.html#method-i-round

This is quite correct when you realize that `format_number` returns a number, not a string. And of course:

```
irb(main):001:0> 3.14 == 3.14000
=> true
```

cllns:
`format_number` [returns a String](https://github.com/hanami/helpers/blob/83e503078657aaec13d718534311be7074b2d509/lib/hanami/helpers/number_formatting_helper.rb#L31), not a number.

Maybe we should change the implementation to something like: 
```
sprintf('%.4f', "3.14") #=> "3.1400"
```
?

TiteiKo:
@cllns Yep. One detail left: in case of an integer, user friendly use would be to have a default precision to 0, to avoid having `format_number(1000) == "1,000.00"` instead of `"1,000"`

PR #85, opened 2016-11-18
Fix number formatting precision

katafrakt:
This fixes https://github.com/hanami/helpers/issues/77

New method is only applied when precision is explicitly passed in options. It does not use default precision for that, because it would be backwards incompatible.

I didn't include solution for secon issue from #77, as I belier pull requests should be atomic. The issue was:

> One detail left: in case of an integer, user friendly use would be to have a default precision to 0, to avoid having format_number(1000) == "1,000.00" instead of "1,000"

### DIFF
--- a/lib/hanami/helpers/number_formatting_helper.rb
+++ b/lib/hanami/helpers/number_formatting_helper.rb
@@ -183,7 +183,7 @@ module Hanami
         def to_str
           number = to_number
           if precision_requested_explicity?
-            "%.#{precision}f" % number
+            Kernel.format("%.#{precision}f", number)
           else
             number.to_s
           end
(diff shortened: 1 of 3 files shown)


jodosha:
@katafrakt This looks great, thank you.

jodosha:
@TiteiKo Please review too.

TiteiKo:
@katafrakt From what I see, your PR is also fulfilling the detail on integer precision being 0 when no precision is explicitly given. 

Looks good to me, thanks 👍

joneslee85:
👍

jodosha:
Thank you all, merging!

katafrakt:
I have just noticed a shameful typo in my code, where method is named `precision_requested_explicity?` instead of `precision_requested_explicitly?`. Can some maintainer fix it maybe, please? 😉

PR #86, opened 2016-11-20
Fix FormBuilder populating form values

### LINKED ISSUE #84: FormBuilder not filling the input fields
Given a form with passed values: 

```ruby
<%= form_for :entity, routes.entities_path, values: { entity: Entity.new(field: 1) }
text_field :field
end %>
```

The text_field will not be filled at all. I tracked the problem down to the `Values#_get_from_values` [message](https://github.com/hanami/helpers/blob/master/lib/hanami/helpers/form_helper/values.rb#L44).

Now that entities are implicitly serializable to hashes (as they respond to `to_hash`), they become hashes during the `Values#initialize` and therefore the `public_send(k)` will always return `nil`.

myabc:
Apparent regression in Hanami 0.9.x series.

Fixes #84

### DIFF
--- a/hanami-helpers.gemspec
+++ b/hanami-helpers.gemspec
@@ -25,4 +25,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler',  '~> 1.6'
   spec.add_development_dependency 'rake',     '~> 11'
   spec.add_development_dependency 'minitest', '~> 5.5'
+
+  spec.add_development_dependency 'dry-struct', '~> 0.1'
 end
--- a/lib/hanami/helpers/form_helper/values.rb
+++ b/lib/hanami/helpers/form_helper/values.rb
@@ -18,7 +18,7 @@ module Hanami
         # @since 0.2.0
         # @api private
         def initialize(values, params)
-          @values = Utils::Hash.new(values || {}).symbolize!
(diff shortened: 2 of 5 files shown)


myabc:
@jodosha this appears to me to be the simplest solution. I do not really see the value in teaching `Values#_get_from_values` to understand hashes.

/ @mereghost

mereghost:
Yep this could solve the problem, but I'm not sure it's the way everything else is going. Even `Interactor`s are [returning simple hashes](https://github.com/hanami/utils/blob/master/lib/hanami/interactor.rb#L136) since the changes on `Model` to become frozen and implicit convertible to a hash. That would solve the issue with the form builder but leave the whole thing a bit inconsistent.

OTOH, if the forms only operated on Hash-like objects (Params & Utils::Hash) the internals could be simplified and testing would become simpler too.

myabc:
@mereghost Alternative solution based on your feedback here: https://github.com/hanami/helpers/pull/87

> OTOH, if the forms only operated on Hash-like objects (Params & Utils::Hash) the internals could be simplified and testing would become simpler too.

mereghost:
At the same time, the time to break compatibility is right now, before we move to 1.0.

myabc:
Closing in favour of #87

PR #87, opened 2016-11-21
Fix FormBuilder populating form values (alternative)

### LINKED ISSUE #84: FormBuilder not filling the input fields

myabc:
Apparent regression in Hanami 0.9.x series.

Alternative solution to #86

Fixes #84

### DIFF
--- a/hanami-helpers.gemspec
+++ b/hanami-helpers.gemspec
@@ -25,4 +25,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler',  '~> 1.6'
   spec.add_development_dependency 'rake',     '~> 11'
   spec.add_development_dependency 'minitest', '~> 5.5'
+
+  spec.add_development_dependency 'dry-struct', '~> 0.1'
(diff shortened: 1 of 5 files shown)


joneslee85:
I personally 👍 this to #84

jodosha:
@myabc Yes, I like it too. 👍  Thanks Alex!

nessur:
this is a small fix, can you release a 0.5.1 so that the form helper can use Hanami::Entity models?

PR #89, opened 2016-11-28
Radiobuttons and selects values that are not strings

ksenia-zalesnaya:
Extracted from other [PR](https://github.com/hanami/helpers/pull/73). Tests state the problem. I don't know whether `to_s` is best solution, anyone has any ideas?

### DIFF
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -311,6 +311,18 @@ describe Hanami::Helpers::FormHelper do
           actual.must_include %(<input type="checkbox" name="book[free_shipping]" id="book-free-shipping" value="true" checked="checked">)
         end
       end
+
+      describe 'checked_value is boolean' do
+        let(:params) { Hash[book: { free_shipping: 'true' }] }
+
+        it "renders with 'checked' attribute" do
+          actual = view.form_for(:book, action) do
+            check_box :free_shipping, checked_value: true
+          end.to_s
+
+          actual.must_include %(<input type="checkbox" name="book[free_shipping]" id="book-free-shipping" value="true" checked="checked">)
+        end
+      end
     end
 
     describe 'automatic values' do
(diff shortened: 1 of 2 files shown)


jodosha:
@TiteiKo Hi. Both this PR and the other one by @ksenia-zalesnaya (#73) have git conflicts because we merged your #81.

Would you be so kind to review the code and make sure that both the fixes are merged? Thank you all! 😄

test/form_helper_test.rb
```diff
@@ -1118,16 +1118,32 @@
...
+            radio_button :category, 'Non-Fiction'
+          end.to_s
+
+          actual.must_include %(<input type="radio" name="book[category]" value="Fiction">\n<input type="radio" name="book[category]" value="Non-Fiction" checked="checked">)
+        end
+      end
+
+      describe 'decimal value' do
+        let(:params) { Hash[book: { price: val }] }
+        let(:val)    { '20.0' }
```
TiteiKo: To test on a decimal value, shouldn't the val be `20.0` and not `'20.0'`?

test/form_helper_test.rb
```diff
@@ -1256,15 +1272,31 @@
...
 
-          actual.must_include %(<select name="book[store]" id="book-store">\n<option>Select a store</option>\n<option value="it" selected="selected">Italy</option>\n<option value="us">United States</option>\n</select>)
+            actual.must_include %(<select name="book[store]" id="book-store">\n<option>Select a store</option>\n<option value="it" selected="selected">Italy</option>\n<option value="us">United States</option>\n</select>)
+          end
+        end
+
+        describe 'integer values' do
+          let(:values) { Hash['Brave new world' => 1, 'Solaris' => 2] }
+          let(:params) { Hash[bookshelf: { book: val }] }
+          let(:val)    { '1' }
```
TiteiKo: Same question as above

test/form_helper_test.rb
TiteiKo: As the params sent to the form are the one before cast, this PR is needed :)

ksenia-zalesnaya: So they should be strings, isn't it?

jodosha:
@ksenia-zalesnaya @TiteiKo is this PR ready? Can we merge it? ✨  😄

test/form_helper_test.rb
TiteiKo: They can either be string or others, so the test is fine 👍

TiteiKo:
@ksenia-zalesnaya Thanks a lot for this PR ⭐️

PR #93, opened 2016-12-18
Fix typo in select documentation

davydovanton:
You can see that in documentation we call `stores` instead `store`
http://www.rubydoc.info/gems/hanami-helpers/Hanami/Helpers/FormHelper/FormBuilder#select-instance_method

/cc @hanami/documentation


davydovanton:
thanks for merge 👍

PR #95, opened 2017-01-10
Escape input values

kwando:
Input values can come from params, hence we can end up with a broken form on malicious input


jodosha:
@kwando Thanks for this PR, but we just merged #96 cause it includes tests.

PR #96, opened 2017-01-13
Form helper doesn't escape value attribute for values entities

davydovanton:
Closes https://github.com/hanami/hanami/issues/709

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -957,7 +957,9 @@ module Hanami
         # @api private
         # @since 0.2.0
         def _attributes(type, name, attributes)
-          { type: type, name: _input_name(name), id: _input_id(name), value: _value(name) }.merge(attributes)
+          attrs = { type: type, name: _input_name(name), id: _input_id(name), value: _value(name) }.merge!(attributes)
+          attrs[:value] = Hanami::Utils::Escape.html(attrs[:value])
+          attrs
         end
 
         # Input <tt>name</tt> HTML attribute
--- a/test/form_helper_test.rb
+++ b/test/form_helper_test.rb
@@ -1023,6 +1023,28 @@ describe Hanami::Helpers::FormHelper do
       end
     end
 
+    describe 'without values' do
+      let(:book)   { Book.new(title: val) }
+      let(:val)    { '"DDD" Book' }
+
+      it 'renders with value' do
+        actual = view.form_for(:book, action) do
+          text_field :title
+        end.to_s
+
+        actual.must_include %(<input type="text" name="book[title]" id="book-title" value="">)
+      end
+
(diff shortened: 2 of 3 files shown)


lib/hanami/helpers/html_helper/empty_html_node.rb
```diff
@@ -124,7 +124,11 @@ def boolean_attribute(attribute_name, _value)
         end
 
         def attribute(attribute_name, value)
-          %(#{ATTRIBUTES_SEPARATOR}#{attribute_name}="#{value}")
+          if attribute_name.to_sym == :value
+            %(#{ATTRIBUTES_SEPARATOR}#{attribute_name}="#{Hanami::Utils::Escape.html(value)}")
+          else
+            %(#{ATTRIBUTES_SEPARATOR}#{attribute_name}="#{value}")
```
AlfonsoUceda: Shouldn't we escape the value in all cases?

jodosha: Probably yes.

Please also consider to use `html_attribute` instead https://github.com/hanami/utils/blob/ca91c336ffebb43f1632ffd398b4281b3b7315a5/lib/hanami/utils/escape.rb#L453

davydovanton: We have some attributes like a `script`. If we escape this attributes we will have a error in layout. What do you think what we should do with this attributes?

TiteiKo: Could we simply make a list of the attributes that need escaping? (like `value`, `title`, ...)

jodosha: @TiteiKo We can't predict where developers are going to interpolate user's input.

jodosha: Alternatively we can make this change in `_value`: https://github.com/hanami/helpers/blob/8f10dbd7475bd28f6197bd667540577cccad1504/lib/hanami/helpers/form_helper/form_builder.rb#L984

WDYT?

davydovanton: @jodosha good idea 👍

davydovanton:
Okay, I updated `#_attributes` form helper because `#_value` sets default value and we need to escape all values. Also I used `#html_attribute` method instead `#html`.

jodosha:
@davydovanton I don't understand the difference. Why it can't be in `_value`?

davydovanton:
I fixed CI build. 

`_value` used only for the default value. If you have a value attribute in your form, you will redefine this value. That's why I escape value after the merge.

davydovanton:
Also, I checked this in my local app, and all worked correctly 🎉

davydovanton:
@jodosha thanks for help and review 👍

PR #98, opened 2017-02-28
Introduce time helper module with relative time method

davydovanton:
So, I think time helpers will helpful for other developers too.

This changes for 1.1.0 release

/cc @hanami/core @jodosha @cllns

### DIFF
--- a/test/time_helper_test.rb
+++ b/test/time_helper_test.rb
@@ -1,0 +1,140 @@
+require 'test_helper'
+
+describe Hanami::Helpers::TimeHelper do
+  before do
+    @view = TimeView.new
+  end
+
+  let(:date)   { Time.now }
+
+  let(:minute) { 60 }
+  let(:hour)   { 60 * minute }
+  let(:day)    { 24 * hour }
+  let(:week)   { 7  * day }
+  let(:month)  { 4  * week }
+  let(:year)   { 12 * month }
+
+  describe '#in_words' do
+    describe 'with from date' do
+      let(:date_from) { Time.now + 1 * hour }
+      it { @view.created_at(date, date_from).must_equal 'a hour ago' }
+
+      describe 'with both DateTime type' do
+        let(:date) { DateTime.new(2017, 1, 15, 11, 0) }
+        let(:date_from) { DateTime.new(2017, 1, 18, 15, 0) }
+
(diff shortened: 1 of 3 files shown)


TiteiKo:
I'm sorry but I don't think it's a good idea... (and I'm against hard coded english translations)

Displaying time formatted this way server side is generally a bad idea as it removes the possibility to cache the page. Furthermore, the text won't update itself as the user stays on the page. There's a few js libraries that are doing exactly this.

test/time_helper_test.rb
```diff
@@ -0,0 +1,140 @@
...
+  let(:hour)   { 60 * minute }
+  let(:day)    { 24 * hour }
+  let(:week)   { 7  * day }
+  let(:month)  { 4  * week }
+  let(:year)   { 12 * month }
+
+  describe '#in_words' do
+    describe 'with from date' do
+      let(:date_from) { Time.now + 1 * hour }
+      it { @view.created_at(date, date_from).must_equal 'a hour ago' }
```
cllns: should be `an hour` (since "hour" sounds like "our", it's treated as though it begins with a vowel... silly english 🙃)

AlfonsoUceda:
Hi folks, @davydovanton thanks for this PR, I think it's a good idea but let me explain. @TiteiKo, @cllns you are right, there are JS libraries for updating the time in real time but I think it's good provide this helper, why? because when you render the time and the JS load, there may be a gap that you can see a glitch between the server rendering and the client rendering.

@davydovanton why not use I18n for this? people could have an use case for using another language

davydovanton:
@AlfonsoUceda thanks for the feedback 👍 
It's a simple version of helper. I think we need to use small iterations and we can add i18n in future (I don't like long PRs). That's why I (or other developers) can improve (or add new helpers) in future. I mean that this PR is just a push for next improvements

mereghost:
@davydovanton While I can see the benefit of making it an interactive development thing, we are past 1.0, so I think shipping fully functional stuff may be a bit better than something that will probably add more tickets on the short run.

Also time math is complicated specially since daylight saving times are a thing. Not sure if we need to tackle this or just recommend some other gem to deal with it.

jodosha:
Given the team has expressed an opinion against this feature, I'm gonna close this.

However, we need to explore client side libraries, in order to give an answer to the developers that need this feature.

PR #102, opened 2017-03-08
API docs cleanup

jodosha:
```shell
➜ bundle exec yardoc --no-cache --no-save --protected --private --no-api private --markup markdown lib
Files:          10
Modules:         8 (    0 undocumented)
Classes:         3 (    0 undocumented)
Constants:       1 (    0 undocumented)
Attributes:      0 (    0 undocumented)
Methods:        33 (    0 undocumented)
 100.00% documented
```

### DIFF
--- a/lib/hanami/helpers/form_helper.rb
+++ b/lib/hanami/helpers/form_helper.rb
@@ -74,9 +74,8 @@ module Hanami
     #     end
     #   end
     #
-    #   # Corresponding template:
-    #   #
-    #   #  <%= my_form %>
+    #   <!-- use this in the template -->
+    #   <%= my_form %>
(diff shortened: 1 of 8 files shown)


lib/hanami/helpers/form_helper.rb
```diff
@@ -74,9 +74,8 @@ module Helpers
     #     end
     #   end
     #
-    #   # Corresponding template:
-    #   #
-    #   #  <%= my_form %>
+    #   <!-- use this in the template -->
```
davydovanton: Maybe is better to use ruby comment? For example:
```
# in any template file:
<%= my_form %>
```

PR #103, opened 2017-03-08
Update HTML elements

jodosha:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element

### DIFF
--- a/lib/hanami/helpers/html_helper/html_builder.rb
+++ b/lib/hanami/helpers/html_helper/html_builder.rb
@@ -42,6 +42,7 @@ module Hanami
           datalist
           del
           details
+          dialog
           dfn
           div
           dl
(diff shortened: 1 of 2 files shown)


lib/hanami/helpers/html_helper/html_builder.rb
```diff
@@ -61,14 +62,14 @@ class HtmlBuilder # rubocop:disable Metrics/ClassLength
...
           header
+          hgroup
           i
           iframe
           ins
           kbd
           label
           legend
           li
-          link
```
AlfonsoUceda: why remove this?

jodosha: It was defined twice. 

Please read the entire commit message: https://github.com/hanami/helpers/pull/103/commits/6736ed902f45a9a0df2c84cb706299fe192a135d 😉

AlfonsoUceda: ok, in EMPTY_TAGS constant 👍

PR #104, opened 2017-03-08
Update form elements

jodosha:
This PR introduces new form helpers:

  * `image_button` (`<input type="image">`)
  * `range_field` (`<input type="range">`)
  * `search_field` (`<input type="search">`)
  * `url_field` (`<input type="url">`)
  * `tel_field` (`<input type="tel">`)

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -786,6 +786,35 @@ module Hanami
           input _attributes(:number, name, attributes)
         end
 
+        # Range input
+        #
+        # You can also make use of the `max`, `min`, and `step` attributes for
+        # the HTML5 number field.
+        #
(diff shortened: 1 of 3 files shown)


jodosha:
Probably tweaking YARD in some way, but I'm not sure if it's worth the trouble.

jodosha:
@davydovanton @mereghost This is now ready to be reviewed. Can you please have a look? Thanks! 👍

jodosha:
@davydovanton WDYM?

davydovanton:
I mean that some helpers are rarely used but we have code for this and we need to support this code too. That's why I have this skeptic question.

jodosha:
@davydovanton I see. Well, it isn't too much burden to maintain these one-line functions IMO 😉 . We also encourage and support HTML5 specific inputs, that act as validator on the browser side.

PR #107, opened 2017-08-11
properly handle nested field indexes

glaszig:
before this patch

    fields_for_collection :addresses do
      text_field :street
      text_field :city
    end

would generate

    <input type="text" name="delivery[addresses][][street]">
    <input type="text" name="delivery[addresses][][city]">

which, in turn, would be interpreted as

    {
      addresses: [
        { street: '...' },
        { city: '...' },
        ...
      ]
    }

which is not wanted because all fields within the set should relate to one item of the set.

with the proper indexing in the name attribute we instead get

    <input type="text" name="delivery[addresses][0][street]">
    <input type="text" name="delivery[addresses][0][city]">

and proper interpretation

    {
      addresses: [
        { street: '...', city: '...' },
        ...
      ]
    }

with my changes one can actually generate form fields for multiple objects in of a collection and it'll keep the indexes.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -233,8 +233,8 @@ module Hanami
         #   <form action="/deliveries" method="POST" accept-charset="utf-8" id="delivery-form">
         #     <input type="hidden" name="_csrf_token" value="920cd5bfaecc6e58368950e790f2f7b4e5561eeeab230aa1b7de1b1f40ea7d5d">
         #     <input type="text" name="delivery[customer_name]" id="delivery-customer-name" value="">
-        #     <input type="text" name="delivery[addresses][][street]" id="delivery-address-0-street" value="">
-        #     <input type="text" name="delivery[addresses][][street]" id="delivery-address-1-street" value="">
+        #     <input type="text" name="delivery[addresses][0][street]" id="delivery-address-0-street" value="">
+        #     <input type="text" name="delivery[addresses][1][street]" id="delivery-address-1-street" value="">
         #
         #     <button type="submit">Create</button>
         #   </form>
@@ -1444,7 +1444,7 @@ module Hanami
         # @api private
         # @since 0.2.0
         def _attributes(type, name, attributes)
-          attrs = { type: type, name: _displayed_input_name(name), id: _input_id(name), value: _value(name) }
+          attrs = { type: type, name: _input_name(name), id: _input_id(name), value: _value(name) }
           attrs.merge!(attributes)
           attrs[:value] = escape_html(attrs[:value])
           attrs
--- a/spec/integration/form_helper_spec.rb
+++ b/spec/integration/form_helper_spec.rb
@@ -91,8 +91,8 @@ RSpec.describe 'Form helper' do
   describe 'form with nested structures' do
     describe 'first page load' do
       before do
-        @address1 = Address.new(street: '5th Ave')
-        @address2 = Address.new(street: '4th Ave')
+        @address1 = Address.new(street: '5th Ave', city: 'New York')
+        @address2 = Address.new(street: '4th Ave', city: 'Seattle')
         @bill     = Bill.new(id: 1, addresses: [@address1, @address2])
         @params   = BillParams.new({})
         @session  = Session.new(_csrf_token: 's14')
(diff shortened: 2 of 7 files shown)


cllns:
Adding @TiteiKo for review b/c I know she did work on collections in views :)

glaszig:
thanks. while at it, can you folks explain the purpose of `#_displayed_input_name` to me? i see what it does but i can't fathom why or what it's for.

glaszig:
another issue closely related arises.

when you submit the example input fields

```html
<input type="text" name="delivery[addresses][0][street]" value="Sin Street">
<input type="text" name="delivery[addresses][0][city]" value="Sin City">
```

you'll end up with the following payload

```rb
payload = "delivery[addresses][0][street]=Sin+Street&delivery[addresses][0][city]=Sin+City"
```

which Rack will parse as

```rb
rack_params = Rack::Utils.parse_nested_query payload
# => {"delivery"=>{"addresses"=>{"0"=>{"street"=>"Sin Street", "city"=>"Sin City"}}}}
```

and Hanami will _symbolize_ to

```rb
params = Hanami::Action::BaseParams.new({"delivery"=>{"addresses"=>{"0"=>{"street"=>"Sin Street", "city"=>"Sin City"}}}})
# => #<Hanami::Action::BaseParams:0x007fe882487b60
# @env={"delivery"=>{"addresses"=>{"0"=>{"street"=>"Sin Street", "city"=>"Sin City"}}}},
# @params={:delivery=>{:addresses=>{:"0"=>{:street=>"Sin Street", :city=>"Sin City"}}}},
# @raw={"delivery"=>{"addresses"=>{"0"=>{"street"=>"Sin Street", "city"=>"Sin City"}}}}>
```

`Hanami::Helpers::FormHelper::Values` will then try to get the input tag value in [`#_get_from_params`](https://github.com/hanami/helpers/blob/5bb0602a64c62c292f45c0651745199096f803b2/lib/hanami/helpers/form_helper/values.rb#L38):

```rb
# @since 0.5.0
# @api private
def _get_from_params(*keys)
  keys.map! { |key| key.to_s =~ /\A\d+\z/ ? key.to_s.to_i : key }
  @params.dig(*keys)
end
```

**here's the problem**: the first line of this method assumes any key that _looks like a digit_ to be an integer and converts it. the following call to `#dig` on the above `BaseParams` instance will then fail because the `addresses` collection is a hash with positional keys as opposed to zero-index array keys:

```rb
params.dig :delivery, :addresses, 0, :street
# => nil
```

```rb
params.dig :delivery, :addresses, "0".to_sym, :street
# => "Sin Street"
```

also have a failing test case ready for this. i'm not sure what to do now. recursively type-checking every value of the params hash and only then decide whether to `#dig` with integer or symbol? seems kind of inelegant.

input welcome. thank you.

glaszig:
working with the form helper even more, i just noticed that the `#select` helper method properly generates integer-indexed fields when scoped in a collection by [calling `#_input_name` instead of `_displayed_input_name`](https://github.com/hanami/helpers/blob/0cf7e87eb3ae5abb52f1636af3da6b1855ada19e/lib/hanami/helpers/form_helper/form_builder.rb#L1539).

TiteiKo:
:-1:

The way Rack interprets the data is correct, and the names of the fields are already correct.

What we want is an **array** of data schemas, not a **hash** of data schemas.

Giving a key (0, 1, ...) to the element implies that the structure is:

```ruby
Hash[
  0: Hash{],
  1, Hash[]
]
```
instead of
```ruby
[Hash[], Hash[]]
```

How then could rack (or hanami) differenciate between someone who wants an array of hashes and someone who wants a Hash with an id as key?

cllns:
@glaszig Thanks for all this hard work. This does seems a little screwy, but as you saw, and @TiteiKo confirmed, you run into issues trying to do this the way you suggested (which is more intuitive, I agree). Especially important is that we don't want to cause some incompatibility with what people expect from Rack parsing, even if it would be more friendly in some ways.

Does that make sense? I'm going to close this issue, but thanks again for putting so much time and thought into this. 🌸

glaszig:
@TiteiKo @cllns how do you suggest should we construct a form with `fields_for_collection` that generates proper form fields for each attribute of an object in the collection?

```rb
fields_for_collection :addresses do
  text_field :street
  text_field :city
end
```

will result in 

```
[Hash[street: "..."], Hash[city: "..."]]
```

when what we really want is (my pr contained an actual test case for the problem)

```
[Hash[street: "...", city: "..."], ...]
```

glaszig:
and what's the point of the `#select` helper generating a `name` attribute different from the other helpers?

TiteiKo:
@glaszig Your assumption that it will result in `[Hash[street: "..."], Hash[city: "..."]]` is wrong, take a look at [this article](http://codefol.io/posts/How-Does-Rack-Parse-Query-Params-With-parse-nested-query) for more info on how Rack works.

Regarding for the `#select` helper generating a different `name` attribute, that's not normal and I'll fix this right away

PR #109, opened 2017-08-29
Form helper: disable option for select prompt

TiteiKo:
Prompt option shouldn't be enabled in generated select

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1239,7 +1239,7 @@ module Hanami
           selected   = options.delete(:selected)
 
           super(attributes) do
-            option(prompt, disabled: DISABLED) unless prompt.nil?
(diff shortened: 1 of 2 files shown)


TiteiKo:
Please don't merge, I want to add the selected attribute to the prompt if there's nothing selected

cllns:
👍 good find and good work!

jodosha:
@TiteiKo Any progress on this enhancement? Would you like to see it in 1.1.0?

AlfonsoUceda:
ping @TiteiKo

AlfonsoUceda:
Folks, after checking this PR in deep I'm against this change, disabled attribute shouldn't be set by default for prompt option, the use should do this.

For example, usually we use prompt for default selected option and we select some valid option, after some time we want to disable that change, selecting again our prompt option.

@jodosha @TiteiKo I'm going to close it, if you think I'm wrong, please comment this PR.

Issue #118, opened 2017-11-07
The `select` method expects an inverted hash for its `values` parameter


malin-as:
Hi,

When providing values to `select`, the code expects a hash such as:
```ruby
{ text: id, text: id, … }
```
I think that it's a very surprising way of doing things, and actually even the documentation is confused (cf http://www.rubydoc.info/gems/hanami-helpers/Hanami/Helpers/FormHelper/FormBuilder#select-instance_method ):
```ruby
values = Hash['Italy' => 'it', 'United States' => 'us']
```
appears in the first two examples, while:
```ruby
values = Hash['it' => 'Italy', 'us' => 'United States']
```
appears in the subsequent six ones.

While the easy fix is to update the documentation, I think a breaking change could be considered instead, to have the hashes in the "right" order.

cllns:
@malin-as Thanks for reporting this. I agree the values should be the key, and the text should be the value. You're right though that this is a breaking change. We should consider doing this for Hanami 2.0.

In the meantime, could you open a PR fixing those comments, to reflect how this helper actually works?

AlfonsoUceda:
I agree with this change, for me they key in the hash should be de option's value and the value in the hash the value to display in the option.

AlfonsoUceda:
cc @jodosha

jodosha:
@malin-as Thanks for this suggestion. It used to be like that in the past, but we changed it as of `v0.4.0`. Please read why here: https://github.com/hanami/helpers/pull/61

PR #119, opened 2017-11-08
Fix the documentation of the `select` helper method.

malin-as:
The hash parameter used to determine the contents of a `select` node has an unexpected format, cf issue #118. This commit makes the documentation consistent with how the `select` method currently works.


cllns:
Thank you!

jodosha:
@cllns Please remember to add milestones to PRs/issues. Thanks.

PR #120, opened 2017-11-09
Fix select with nil option

### LINKED ISSUE #115: `select` behaves in an unexpected way with `nil` in options
When using `nil` as a value in the options for a `select`, unexpected things are happening.

Example:
```erb
<%=
  thing = OpenStruct.new(key: 'foo')
  form_for :thing, '/books/1', method: :put, values: {thing: thing} do
    select_options = [
      ['Foo', 'foo'],
      ['Bar', 'bar'],
      ['N/A', nil]
    ]
    select :key, select_options
    submit 'Save'
  end
%>
```

Results in:
```html
<form action="/books/1" method="POST" accept-charset="utf-8" id="thing-form">
<input type="hidden" name="_method" value="PUT">
<select name="thing[key]" id="thing-key">
<option value="foo" selected="selected">Foo</option>
<option value="bar">Bar</option>
<option value="" selected="selected">N/A</option>
</select>
<button type="submit">Save</button>
</form>
```

Note that **two** values are `selected`: The correct one (Foo) as well as the value for `nil` (N/A).

Visually, this results in the wrong value being selected in the UI:
web

AlfonsoUceda:
This PR fixes #115 

@nerdinand can you take a look this fix?

After merge, please cherry pick to master.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1206,8 +1206,9 @@ module Hanami
           super(attributes) do
             option(prompt) unless prompt.nil?
 
+            already_selected = nil
             values.each do |content, value|
-              if _select_option_selected?(value, selected, _value(name), attributes[:multiple])
+              if (attributes[:multiple] || !already_selected) && already_selected = _select_option_selected?(value, selected, _value(name), attributes[:multiple])
                 option(content, { value: value, selected: SELECTED }.merge(options))
               else
                 option(content, { value: value }.merge(options))
@@ -1547,8 +1548,10 @@ module Hanami
         # rubocop:disable Metrics/CyclomaticComplexity
         # rubocop:disable Metrics/PerceivedComplexity
         def _select_option_selected?(value, selected, input_value, multiple)
-          value == selected || (multiple && (selected.is_a?(Array) && selected.include?(value))) ||
-            value.to_s == input_value.to_s || (multiple && (input_value.is_a?(Array) && input_value.include?(value)))
+          value == selected ||
+            (multiple && (selected.is_a?(Array) && selected.include?(value))) ||
+            value.to_s == input_value.to_s ||
+            (multiple && (input_value.is_a?(Array) && input_value.include?(value)))
         end
         # rubocop:enable Metrics/PerceivedComplexity
         # rubocop:enable Metrics/CyclomaticComplexity
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1208,7 +1208,7 @@ module Hanami
 
             already_selected = nil
             values.each do |content, value|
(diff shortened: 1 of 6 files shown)


spec/unit/hanami/helpers/form_helper_spec.rb
```diff
@@ -2427,6 +2427,34 @@
         expect(actual).to include(%(<select name="book[store]" id="book-store">\n<option value="it" selected="selected">Italy</option>\n<option value="us">United States</option>\n</select>))
       end
     end
+
+    describe 'with nil as a value' do
+      let(:option_values) { Hash['Italy' => 'it', 'United States' => 'us', 'N/A' => nil] }
+
+      it "doesn't set options as selected" do
```
nerdinand: This should probably be `it 'sets nil option as selected'`

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -1206,8 +1206,9 @@ def select(name, values, attributes = {}) # rubocop:disable Metrics/AbcSize, Met
           super(attributes) do
             option(prompt) unless prompt.nil?
 
+            already_selected = nil
             values.each do |content, value|
-              if _select_option_selected?(value, selected, _value(name), attributes[:multiple])
+              if (attributes[:multiple] || !already_selected) && already_selected = _select_option_selected?(value, selected, _value(name), attributes[:multiple])
```
davydovanton: WDYT, maybe will better (for understanding) replace this condition to separate variable or method?

spec/unit/hanami/helpers/form_helper_spec.rb
```diff
@@ -2427,6 +2427,34 @@
...
+        actual = view.form_for(:book, action) do
+          select :store, option_values
+        end.to_s
+
+        expect(actual).to include(%(<select name="book[store]" id="book-store">\n<option value="it">Italy</option>\n<option value="us">United States</option>\n<option value="" selected="selected">N&#x2F;A</option>\n</select>))
+      end
+
+      it "set as selected the option with nil value" do
```
nerdinand: This has the same outcome as the previous case. But I guess it's fine to test them both...

lib/hanami/helpers/form_helper/form_builder.rb
nerdinand: I personally don't like assignments in if statements, but I'm not sure what your policy is...

nerdinand (commented):
Thanks for fixing this!

lib/hanami/helpers/form_helper/form_builder.rb
AlfonsoUceda: I could create another line, let see what say others ;)

spec/unit/hanami/helpers/form_helper_spec.rb
AlfonsoUceda: ups, you're right, let me fix it.

davydovanton (approved):

jodosha (approved):
LGTM 👍 Thanks @AlfonsoUceda

PR #125, opened 2017-11-28
Ensure FormBuilder#select to respect "selected" option

### LINKED ISSUE #124: Placeholder option tag in select gets selected by default
With regards to hanami-helpers 1.1.0, the following code:

```ruby
require 'hanami-helpers'

include Hanami::Helpers

def params
  {}
end

h = html do
  form_for :test, '/test' do
    values = { 'Italy' => 'it',
               '---' => '' }
    select :country, values
  end
end

puts h
```
produces:
```html
<form action="/test" method="POST" accept-charset="utf-8" id="test-form">
    <select name="test[country]" id="test-country">
        <option value="it">Italy</option>
        <option value="" selected="selected">---</option>
    </select>
</form>
```
The placeholder is marked as selected when it should not. This also happens with a value of `nil` (i.e. `'---' => nil`).

jodosha:
Ensure `FormBuilder#select` to respect `selected:` option:

```erb
<%=
  form_for :test, '/test' do
    values = { "Italy" => "it", "---" => "" }
    select :country, values, options: { selected: 'none' }
  end
%>
```

```html
<form action="/test" method="POST" accept-charset="utf-8" id="test-form">
<select name="test[country]" id="test-country">
<option>---</option>
<option value="it">Italy</option>
</select>
</form>
```

---

Ref #120
Fixes #124

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1550,7 +1550,7 @@ module Hanami
         def _select_option_selected?(value, selected, input_value, multiple)
           value == selected ||
             (multiple && (selected.is_a?(Array) && selected.include?(value))) ||
-            value.to_s == input_value.to_s ||
+            (!input_value.nil? && (value.to_s == input_value.to_s)) ||
             (multiple && (input_value.is_a?(Array) && input_value.include?(value)))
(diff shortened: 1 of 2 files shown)


AlfonsoUceda (approved):
LGTM

PR #128, opened 2018-02-17
Join attribute values with space, for arrays

cllns:
Without this change, `class: [:ui, :form]` renders, in HTML, as:

```html
class="["ui", "form"]
```
which is not what a developer would expect.

With this change, `class: [:ui, :form]` renders, in HTML, as:

```html
 class="ui form"
```
 as one would expect.

Note: I implemented this for _all_ attributes, not just `class`. I'm not sure if that makes sense? I can't think of other attributes that would act this way.

If we only want to do this for `class`, it actually makes the code more perform better, since we don't have to go through the process of iterating through the Hash and creating a new copy of it.

I also only wrote the spec for `input`, but we should have it for all the attributes. Just wanted to make sure others were interested in this change before I went through the manual work for doing that? (Or can we add some spec elsewhere?)

What do you think @hanami/core and others?

### DIFF
--- a/lib/hanami/helpers/html_helper/empty_html_node.rb
+++ b/lib/hanami/helpers/html_helper/empty_html_node.rb
@@ -78,7 +78,7 @@ module Hanami
         # @api private
         def initialize(name, attributes)
           @name       = name
-          @attributes = attributes
+          @attributes = htmlified_array_attributes(attributes)
         end
 
         # Resolve and return the output
@@ -128,6 +128,21 @@ module Hanami
         # @api private
         def attribute(attribute_name, value)
           %(#{ATTRIBUTES_SEPARATOR}#{attribute_name}="#{value}")
+        end
+
+        # Changes any attributes values that are arrays into a string joined by
+        # a space, as HTML expects it.
+        #
+        # @param attributes [Hash] the attributes with values to be HTMLified
+        #
+        # @return [Hash] the attributes transformed into HTML friendly values, i.e space separated strings
+        #
+        # @since x.x.x
+        def htmlified_array_attributes(attributes)
+          attributes&.inject({}) do |attrs, (key, value)|
+            attrs[key] = value.is_a?(Array) ? value.join(" ") : value
+            attrs
+          end
         end
       end
     end
--- a/lib/hanami/helpers/html_helper/empty_html_node.rb
+++ b/lib/hanami/helpers/html_helper/empty_html_node.rb
@@ -140,7 +140,7 @@ module Hanami
         # @since x.x.x
         def htmlified_array_attributes(attributes)
           attributes&.inject({}) do |attrs, (key, value)|
-            attrs[key] = value.is_a?(Array) ? value.join(" ") : value
+            attrs[key] = value.is_a?(Array) ? value.join(ATTRIBUTES_SEPARATOR) : value
             attrs
           end
         end
--- a/lib/hanami/helpers/html_helper/empty_html_node.rb
+++ b/lib/hanami/helpers/html_helper/empty_html_node.rb
@@ -78,7 +78,7 @@ module Hanami
         # @api private
         def initialize(name, attributes)
           @name       = name
-          @attributes = htmlified_array_attributes(attributes)
+          @attributes = prepare_html_attributes(attributes)
         end
 
         # Resolve and return the output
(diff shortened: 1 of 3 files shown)


lib/hanami/helpers/html_helper/empty_html_node.rb
```diff
@@ -129,6 +129,21 @@ def boolean_attribute(attribute_name, _value)
...
+        # Changes any attributes values that are arrays into a string joined by
+        # a space, as HTML expects it.
+        #
+        # @param attributes [Hash] the attributes with values to be HTMLified
+        #
+        # @return [Hash] the attributes transformed into HTML friendly values, i.e space separated strings
+        #
+        # @since x.x.x
+        def htmlified_array_attributes(attributes)
+          attributes&.inject({}) do |attrs, (key, value)|
```
cllns: There are a number of ways to do this, including `.map.to_h`, but I chose what I think is the most obvious one.

Ruby 2.4 has `Hash#transform_values`, which we can leverage once we stop supporting Ruby 2.3. (Should I add that as a `TODO:` comment?)

lib/hanami/helpers/html_helper/empty_html_node.rb
```diff
@@ -78,7 +78,7 @@ class EmptyHtmlNode
         # @api private
         def initialize(name, attributes)
           @name       = name
-          @attributes = attributes
+          @attributes = htmlified_array_attributes(attributes)
```
cllns: Not married to this name. It's kind of awkward but I wanted it to be specific?

spec/unit/hanami/helpers/html_helper/html_builder_spec.rb
```diff
@@ -245,6 +245,13 @@
       result = @builder.input('required' => true, 'value' => 'Title "book"', 'something' => 'bar').to_s
       expect(result).to eq('<input required="required" value="Title "book"" something="bar">')
     end
+
+    # rubocop:disable Style/SymbolArray
```
cllns: Rubocop wants us to use `%I[]` or `%i[]` here, but that makes it less obvious what's going on, if you're not familiar with that synatx. I can do that instead, here, if others think it's a good idea.

AlfonsoUceda:
@cllns one thing, why not use this? `class: 'ui form'`, I think it is easy and simple.

cllns:
@AlfonsoUceda That is what I did instead. People are free to use that, which is what HTML expects. 

I consider a space-separated string to be an implementation detail of how HTML is implemented. Ruby doesn't expect a space-separated string in order to form a list. I'd like to write my classes in a more Ruby way :) What do you think? It's a little pedantic but I think this is a good enhancement.

lib/hanami/helpers/html_helper/empty_html_node.rb
```diff
@@ -129,6 +129,21 @@ def boolean_attribute(attribute_name, _value)
...
+        # a space, as HTML expects it.
+        #
+        # @param attributes [Hash] the attributes with values to be HTMLified
+        #
+        # @return [Hash] the attributes transformed into HTML friendly values, i.e space separated strings
+        #
+        # @since x.x.x
+        def htmlified_array_attributes(attributes)
+          attributes&.inject({}) do |attrs, (key, value)|
+            attrs[key] = value.is_a?(Array) ? value.join(" ") : value
```
jodosha: @cllns This should use `ATTRIBUTES_SEPARATOR`:

```ruby
value.join(ATTRIBUTES_SEPARATOR)
```

spec/unit/hanami/helpers/html_helper/html_builder_spec.rb
```diff
@@ -245,6 +245,13 @@
...
+    it 'renders attribute with array value as joined with string' do
+      result = @builder.input('class' => [:ui, :form]).to_s
+      expect(result).to eq('<input class="ui form">')
+    end
```
jodosha: @cllns I'd add another spec also to cover `HtmlNode`. For now the implementation of this PR brings the benefits of this enhancement because `HtmlNode` inherits from `EmptyHtmlNode`, but if that hierarchy changes, we may loose track of this feature for `HtmlNode`, unless we add a test.

cllns: :D good call. It actually didn't work for non-empty HTML nodes 😅 . `HtmlNode` doesn't call `super`, so I had to re-do the call to `htmlified_array_attributes`.

PR #132, opened 2018-03-26
fix select with values and nil options

### LINKED ISSUE #131: select: selected option depends on order of options
I just discovered this while upgrading to hanami 1.1.
My tests were failing, claiming the expected option was not selected in a `<select>` element. The same test passed in hanami 1.0. I dug around a bit. Apparently the fix in https://github.com/hanami/helpers/commit/c0043de2d65ef0306edfd7af30914653f8e7f4f9 introduced a new bug:

If your options contain a `nil` option, it must be the last option. Otherwise the rendered selection will not be correct if anything but `nil` is the value of your attribute.

Example:

In my view I have:

```haml
- foo = OpenStruct.new(bar: 'a')

= form_for(:foo, '', values: {foo: foo}) do
  - select :bar, [['a', 'a'], ['b', 'b'], ['nil option', nil]]

= form_for(:foo, '', values: {foo: foo}) do
  - select :bar, [['nil option', nil], ['a', 'a'], ['b', 'b']]
```

This leads to the HTML:
```html
<form action="" method="POST" accept-charset="utf-8" id="foo-form">
<input name="_csrf_token" value="0ed6c40b814d0f394d6221746609fc6bcd4eed530a6c773ce39ad63260835b95" type="hidden">
<select name="foo[bar]" id="foo-bar">
<option value="a" selected="selected">a</option>
<option value="b">b</option>
<option value="">nil option</option>
</select>
</form>
<form action="" method="POST" accept-charset="utf-8" id="foo-form">
<input name="_csrf_token" value="0ed6c40b814d0f394d6221746609fc6bcd4eed530a6c773ce39ad63260835b95" type="hidden">
<select name="foo[bar]" id="foo-bar">
<option value="" selected="selected">nil option</option>
<option value="a">a</option>
<option value="b">b</option>
</select>
</form>
```

Which consequently renders as:

In this example, both selects should show 'a' as the selected option.

GustavoCaso:
Fixes #131

@nerdinand can you take a look this fix?

@jodosha I was playing with this bug and decided to make the first contribution 🎉 

The bug was originated because of the `selected` been set as `nil` and the first value been `nil` as well.

If we are happy with the end result I would like to refactor `_select_option_selected?` before merging.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -81,11 +81,11 @@ module Hanami
 
         # Instantiate a form builder
         #
-        # @overload initialize(form, attributes, params, &blk)
+        # @overload initialize(form, attributes, context, &blk)
         #   Top level form
         #   @param form [Hanami::Helpers:FormHelper::Form] the form
         #   @param attributes [::Hash] a set of HTML attributes
-        #   @param params [Hanami::Action::Params] request params
+        #   @param context [Hanami::Helpers::FormHelper]
         #   @param blk [Proc] a block that describes the contents of the form
         #
         # @overload initialize(form, attributes, params, &blk)
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1225,10 +1225,11 @@ module Hanami
         #     <option value="zw">Zimbabwe</option>
         #   </select>
         def select(name, values, attributes = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-          options    = attributes.delete(:options) { {} }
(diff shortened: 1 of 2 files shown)


nerdinand:
@GustavoCaso thanks so much for fixing this! LGTM, but I agree `_select_option_selected?` is ripe for a refactoring...

GustavoCaso:
Well travis is actually complaining so what better excuse that this one 😄

nerdinand:
@GustavoCaso cool! I think you could even argue for extracting select into its own class though... WDYT?

GustavoCaso:
@nerdinand I totally agree with you. I will extract it to it's own class.

nerdinand:
@GustavoCaso @jodosha I guess the proposed refactoring of `select` into its own class will be done in another PR?

GustavoCaso:
@nerdinand Once the new release is out I will focus on the refactor 😄

PR #135, opened 2018-04-04
Fix form_for signature according to original intent, API docs, and guides

### LINKED ISSUE #134: form_for with only a form object raises an error
[Example in docs](https://github.com/hanami/helpers/blob/fdce06d5123a427b21c37e55f5a6a3fd9943b7d9/lib/hanami/helpers/form_helper.rb#L150-L181) shows how `form_for` can be called with a form object.

However, [method signature](https://github.com/hanami/helpers/blob/master/lib/hanami/helpers/form_helper.rb#L409) looks like this:

```ruby
def form_for(name, url, options = {}, &blk)
```

Note that it has two mandatory positional arguments. This makes it impossible to call the method with only a form object (i.e. without options).

Here's the exception when `form_for` is called only with a form object (`form_for form_object do`):

```
ArgumentError: wrong number of arguments (given 1, expected 2..3)

[...]/hanami-helpers-1.1.1/lib/hanami/helpers/form_helper.rb:409:in `form_for'
```

jodosha:
Fixes #134 

---

Note: the problem wasn't caught until now because the other fixtures are always using more than one argument for `form_for`.

```erb
<%=
  form_for(form, class: "form-horizontal") do
    # ...
  end
%>
```

To test properly this signature, I added `spec/support/fixtures/templates/settings/edit.html.erb`, which only uses `form_for(form)` (one argument).

### DIFF
--- a/lib/hanami/helpers/form_helper.rb
+++ b/lib/hanami/helpers/form_helper.rb
@@ -229,7 +229,7 @@ module Hanami
       #   @option options [Hash] :values An optional payload of objects to pass
       #   @param blk [Proc] A block that describes the contents of the form
       #
-      # @overload form_for(form, attributes, &blk)
+      # @overload form_for(form, attributes = {}, &blk)
       #   Use Form
       #   @param form [Hanami::Helpers::FormHelper::Form] a form object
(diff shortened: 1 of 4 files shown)


AlfonsoUceda (approved):
👍

Issue #114, opened 2017-10-24
JRuby ConcurrencyError with tag block form


nerdinand:
In my code I'm trying to use the block form of `tag`, something like this:
```ruby
  form_for :thing, '/', method: :post do
    tag :label do
      check_box :attribute, value: 'foo'
      text 'bar'
    end
  end
```

When I try to render this with JRuby, I get a ConcurrencyError:

```
ConcurrencyError: Detected invalid array contents due to unsynchronized modifications with concurrent users
	org/jruby/RubyArray.java:2486:in `map'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/html_helper/html_builder.rb:332:in `to_s'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/form_helper/form_builder.rb:140:in `to_s'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/html_helper/html_node.rb:61:in `content'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/form_helper/html_node.rb:49:in `content'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/html_helper/html_node.rb:44:in `to_s'
	org/jruby/RubyArray.java:2486:in `map'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/html_helper/html_builder.rb:332:in `to_s'
	/Users/ferdi/.rbenv/versions/jruby-9.1.13.0/lib/ruby/gems/shared/gems/hanami-helpers-1.0.0/lib/hanami/helpers/form_helper/form_builder.rb:140:in `to_s'
	/Users/ferdi/projects/bookshelf/apps/web/templates/books/edit.html.erb:1:in `block in singleton class'
	org/jruby/RubyBasicObject.java:1691:in `instance_eval'
[...]

PR #136, opened 2018-04-16
Add failing spec to reproduce #114

nerdinand:


AlfonsoUceda:
Thanks @nerdinand, if you want to explore the solution go ahead, if not, I'll try to take a look ;)

nerdinand:
@AlfonsoUceda I tried to figure it out but reproducing it in a test is all I was capable of. I think I don't know the code well enough.

jodosha:
@nerdinand My understanding of #114 was that only JRuby was causing issues for you. From what I can tell from the CI build of this PR, also MRI fails.

Do you know why?

nerdinand:
@jodosha I don't know why, but I noticed it too, see: https://github.com/hanami/helpers/issues/114#issuecomment-381574667

Looks like the two things are unrelated, but I'd still consider them bugs if you agree with the premise of the test...

jodosha:
@nerdinand Maybe they are related..
If I replace `text` with another `radio_button`, I get an infinite loop.

Do you have time to look at this? Please let me know. Thanks.

nerdinand:
@jodosha i currently don't have any time to investigate...

jodosha:
Closed as stale conversation.

PR #137, opened 2018-04-16
FormBuilder: fixing method _select_option_selected?

lucasdfg:
Bug introduced in Hanami 1.2 PR #132

This avoids select element not having a selected option when comparing 5 (String) to 5 (Integer), for example.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1676,7 +1676,7 @@ module Hanami
         # @api private
         def _select_option_selected?(value, selected, input_value, multiple)
           if input_value && selected.nil?
-            value.to_s == input_value
+            value.to_s == input_value.to_s
           else
             (value == selected) ||
               _is_in_selected_values?(multiple, selected, value) ||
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@lucasdfg thanks for this PR and your time, can you add specs to avoid regressions in the future? Thanks!

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -1676,7 +1676,7 @@ def _select_input_name(name, multiple)
         # @api private
         def _select_option_selected?(value, selected, input_value, multiple)
           if input_value && selected.nil?
-            value.to_s == input_value
+            value.to_s == input_value.to_s
```
davydovanton: WDYT about `String(input_value)`?

m45t3r:
@AlfonsoUceda @davydovanton We added the specs, what do you think?

lib/hanami/helpers/form_helper/form_builder.rb
m45t3r: `String(s)` seems to only call `#to_s`, so I don't think this would be beneficial.

jodosha:
@GustavoCaso @nerdinand Can you folks please have a look at this as well? This appears to be related to https://github.com/hanami/helpers/pull/132 Thanks.

nerdinand:
looks good!

jodosha:
@lucasdfg Thanks for this fix. Just merged it. 👍

m45t3r:
@jodosha Can we have a point release with this fix? We are still using a forked repository in our production app, would be great to come back to using a proper release.

jodosha:
@m45t3r Sure thing.

jodosha:
https://rubygems.org/gems/hanami-helpers/versions/1.2.1

PR #138, opened 2018-06-08
Change CSRF meta tag from 'value' to 'content'

cllns:
`value` isn't a valid attribute for a meta tag, according to [the spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta), we want `content` instead.

Discovered this because I was trying to use `hanami-ujs` and it wasn't picking the CSRF meta tag values properly, since it looks for the `content` attribute. I'm using this branc now and it solved the problem. :)

Fixes bug introduced in: #129 

(Intentionally opened against `master` instead of `develop` since this PR is fixing a bug, which can be a patch release.)

### DIFF
--- a/lib/hanami/helpers/form_helper.rb
+++ b/lib/hanami/helpers/form_helper.rb
@@ -457,16 +457,16 @@ module Hanami
       #   <html>
       #     <head>
       #       <!-- ... -->
-      #       <meta name="csrf-param" value="_csrf_token">
-      #       <meta name="csrf-token" value="4a038be85b7603c406dcbfad4b9cdf91ec6ca138ed6441163a07bb0fdfbe25b5">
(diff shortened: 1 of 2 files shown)


cllns:
Just merged it @jodosha. A new release soon would be great but it's not super urgent to me, since I can just use the branch as a source until this is released. (As long as we don't delete the branch :wink:)

jodosha:
@cllns I deleted the branch before to read your comment. Anyway, having that branch around for a specific case is too much. The alternative is to use `master`, but for me is fine to release a new version. 👍

cllns:
Hahah, no worries, I can use `master` then :) Feel free to release whenever it's convenient for you 🌸

jodosha:
@cllns Already released `1.2.2` for you on yesterday. 😉

PR #143, opened 2019-08-07
Adds a disabled attribute to the select option prompt

DannySantos:
Without this attribute Hanami will treat "Select store...", for example, as the value and add it to the params.

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1224,7 +1224,7 @@ module Hanami
         #
         #   <!-- output -->
         #   <select name="book[store]" id="book-store" class="form-control">
-        #     <option>Select a store</option>
+        #     <option disabled="disabled">Select a store</option>
         #     <option value="it">Italy</option>
         #     <option value="us">United States</option>
         #   </select>
@@ -1287,7 +1287,7 @@ module Hanami
           input_value = _value(name)
 
           super(attributes) do
-            option(prompt) unless prompt.nil?
+            option(prompt, disabled: true) if prompt
 
             already_selected = nil
             values.each do |content, value|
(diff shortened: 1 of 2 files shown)


lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -1284,10 +1292,11 @@ def select(name, values, attributes = {}) # rubocop:disable Metrics/AbcSize, Met
           attributes  = { name: _select_input_name(name, multiple), id: _input_id(name) }.merge(attributes)
           prompt      = options.delete(:prompt)
           selected    = options.delete(:selected)
+          disabled    = prompt.nil? ? nil : DISABLED
           input_value = _value(name)
 
           super(attributes) do
-            option(prompt) unless prompt.nil?
+            option(prompt, disabled: disabled) unless prompt.nil?
```
cllns: I think we can remove all the other changes to this file and just do this:
```suggestion
            option(prompt, disabled: true) unless prompt.nil?
```

The `lib/hanami/helpers/html_helper/empty_html_node.rb` file takes care of changing `disabled: true` to be the html `disabled="disabled"` so we don't need to duplicate that logic here.

(Also I know it was already here, but I personally like `if prompt` better, since if prompt is `false`, it's not `nil`, but I think we'd want the same behavior, since both are false-y)

DannySantos: Yeah good shout, I'll make the changes

DannySantos: Ok I made the requested changes and squashed the commits into one, 5 seemed a little excessive for 6 lines of code :)

DannySantos:
@cllns any chance we could get this merged? I have an ugly JavaScript hack in place to get around this that I'd love to take out :)

PR #148, opened 2019-10-14
Add support for block syntax in label helper

### LINKED ISSUE #144: `label` helper doesn't accept block syntax
Similar to #127, but with `label`. I'll make a PR as soon as I can.

Example use case:

```ruby
label for: :pizza do
  text 'Pizza '
  abbr '*', title: 'required', 'aria-label': 'required'
end
```

Which is more accessible than just having `label 'Pizza *', for: :pizza` :slightly_smiling_face: :pizza:

arr-dev:
Closes #144

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -363,16 +363,16 @@ module Hanami
         #   <!-- output -->
         #   <label for="delivery-address-city">City</label>
         #   <input type="text" name="delivery[address][city] id="delivery-address-city" value="">
-        def label(content, attributes = {})
+        def label(content = nil, **attributes, &blk)
           attributes = { for: _for(content, attributes.delete(:for)) }.merge(attributes)
           content    = case content
-                       when String, Hanami::Utils::String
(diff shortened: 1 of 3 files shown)


lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -363,16 +363,21 @@ def fields_for_collection(name, &block)
         #   <!-- output -->
         #   <label for="delivery-address-city">City</label>
         #   <input type="text" name="delivery[address][city] id="delivery-address-city" value="">
-        def label(content, attributes = {})
+        def label(content, attributes = {}, &blk)
+          if content.is_a?(::Hash)
+            attributes = content
+            content = nil
+          end
```
davydovanton: hey, I think you can try to use `**` syntax here:


I mean, instead of this code:
```ruby
        def label(content, attributes = {}, &blk)
           if content.is_a?(::Hash)
             attributes = content
             content = nil
           end
```

you can use this one:
```ruby
        def label(content = nil, **attributes, &blk)
```

WDYT about it?

arr-dev: I'll update the PR later today

PR #166, opened 2020-12-23
Compatibility with hanami-view 2

jodosha:
## What

Compatibility with `hanami-view` 2.

## Notes for the reviewers

Despite the large git diff, production code (`lib/`) changed only of one line to suppress a Ruby warning about non-initialized instance variable.

The most notable changes are in `spec/support/fixtures.rb`. In this file there is a realistic setup for integration specs with the new version of `hanami-view`.

I moved integration specs from `spec/integration/*.rb` to `spec/integration/hanami/helpers/*.rb` for consistency sake with other gems. The way GitHub represents this change may confuse the reviewer: files weren't deleted and we don't have new tests. What you can find under `spec/integration/hanami/helpers/*.rb` are the moved integration specs with RSpec style changes. The testing logic didn't change.

The rest of the changes are just modernizing the RSpec style of the testing code.


spec/integration/hanami/helpers/escape_helper_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+RSpec.describe "Escape helper" do
+  subject do
+    Users::Show.new.call(format: :html, user: user).to_s
+  end
```
timriley, line 6: Seeing this as the subject of a helper test – and having to look elsewhere to see how the helper is actually being used within the template – makes me think that it might be nice to support an "inline template" style of building views. I'll add that as to our Trello to consider for the future :)

spec/support/fixtures.rb
```diff
@@ -297,58 +297,79 @@ class Book < Dry::Struct
 
 User = Struct.new(:name, :website, :snippet)
 
-module TestView
-  def self.included(view)
-    view.class_eval do
-      include Hanami::View
+class TestScope < Hanami::View::Scope
+  include Hanami::Helpers
+end
```
timriley, line 302: I think a more realistic way to set this up for testing would be to include the helpers module inside a context class, rather than a scope class. Doing it this way ensures the helpers are available in _any_ aspect of a view rendering, even if a new scope is used inside a template, e.g.

```erb
<h1>This is my view</h1>

<%= scope(:special_scope).render(:some_partial) %>
```

^ In this example, if the helpers were included in the view's scope, they'd no longer be available inside `_some_partial`, because that's being rendered with a `SpecialScope` as its scope, no longer the view's default scope.

By putting the helpers in the context, they're available within both the top-level template, _and_ the `_some_partial`, even when it's rendered with a custom scope.

Happy for you to consider this for a follow-up PR, however, since what you have here is certainly working for the tests as they currently stand!

timriley (approved):
Great stuff, @jodosha! Nice to see this was so easy to port across :)

I left a note about including the helpers into a context rather than scope, to ensure they're available as widely as possible. I think this would be good to do before we wrap up this particular piece of work, but I'm happy for you to decide whether you'd like to tackle it as part of this initial PR, or do it in a follow-up. Also happy to further discuss the scope/context differences if you'd like some clarification!

